PR #14, opened 2015-01-20
fixed the markdown syntax in README.md

armish:


dkoes:
I'm not sure this is better.  The primary goal of the file is to look good after processing with jsdoc to serve as the [homepage of the documentation](http://3dmol.csb.pitt.edu/doc/index.html).

I assume the problem as that using it as the readme for github results in some ugliness.  The correct solution is probably to have separate files, one for the documentation and one for github.

armish:
oh, OK. I didn't know you were using the same resources -- just wanted to make it look good on GitHub homepage ;)

dkoes:
I've created a separate readme for github so things look nicer.

PR #15, opened 2015-02-02
Modify viewer.html to allow extensions longer than 3 characters.

jus1in:
The mmCIF file format has the extension ".mcif" which is 4 characters, making it not possible to use the viewer before.


dkoes:
I merged this, but it has an infinite loop for some inputs in splitRespectingQuotes which needs to be fixed.

jus1in:
I had only intended to commit viewer.html. I was still working on the other stuff, and I was not intended to be in the pull request.

jus1in:
The second commit also leaves the files in the build folder empty for some reason. Can the merge be reverted?

PR #27, opened 2015-03-11
Remove contents of build/ from source control

jus1in:
I think the build folder should be removed from source control. It is not source code because it is created by the build process, so there is no point in versioning it. All it does is create merge conflicts.

I've created this pull request to discuss this.


jus1in:
You would probably have to change your deployment process to build the code if it doesn't already.

dkoes:
For CDN hosting we need to output of the builds to be in git somehow.  Originally I had a release directory that was distinct from the build directory.  Perhaps if we went back to that system..

jus1in:
Has to be in Git? That's inconvenient.
Wouldn't creating a release directory just mean that there is a file constantly changing in the release directory instead of the build directory?

dkoes:
I meant to have a discussion about this at the group meeting but forgot.. let's discuss at the next meeting.

PR #31, opened 2015-03-17
More vertices in cylinder basisVectors

RacingTadpole:
I thought you might find this generalization of cylinder `basisVectors` helpful. As written, it gives exactly the same result as before, but you can increase `subdivisions` (on line 89) beyond 4 to increase the number of vertices used (and therefore the resolution).  I have not included any way for a user to adjust this.  Nor have I written any tests, but it appears to work for a range of values.

There's also a `sphereQuality` variable to increase the resolution of the spheres, which is useful but a bit of a kludge.  Also please disregard the changes to `.gitignore` and the build of course.


dkoes:
Nice!  Since you left it at the original resolution, there is no performance hit, but it is good to have the option of increasing the resolution.

RacingTadpole:
Great, thanks! After sending that pull request, I realized it would be simpler (and easier to understand) if we define the vertices in a single for-loop over an angle, using sin & cos.  So I may be sending you another pull request soon...

PR #46, opened 2015-04-21
change package version to 1.0.2

RacingTadpole:
Hi, this is a very minor change but it makes it easier to one day install 3Dmol with npm.

Eg. I want to use 3Dmol as an npm dependency with this line in package.json:

      "dependencies": {
          ...
          "3Dmol": "git://github.com/dkoes/3Dmol.js.git"
      },

npm won't install this if the 3Dmol package.json has "version": "1.0" - it needs to be "1.0.2" or similar.

(In fact this change alone isn't sufficient - the build/3Dmol.js file would need to be added back in, and it would need to return an object that npm can use... I might try a pull request for this at some stage too...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,4 @@
 /three.js
 script.map
 my_pdbs/
-build
 doc
--- a/build/3Dmol.js
+++ b/build/3Dmol.js
@@ -1,0 +1,25514 @@
+/*!
+ * jQuery JavaScript Library v1.9.1
+ * http://jquery.com/
+ *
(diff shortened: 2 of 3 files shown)


RacingTadpole:
I closed this because it's an incomplete solution, and I only meant it to be a pull request for the commit 7bd2a94, not for master (which now also d1012de, adding build/3Dmol.js back in).

dkoes:
We switched to putting the checked in built versions in a release directory.  The build directory is now local only.  I have committed the version update.

PR #57, opened 2015-05-28
fix paths for testshape - but it still doesn't work

RacingTadpole:
I wanted to check this example out but it wasn't loading the right scripts. Now it at least does that, although it still doesn't work...


dkoes:
9f9c234 cleans up this example to get it to work as well as fixes a problem with cylinder intersection testing.

RacingTadpole:
Very nice - thanks!

PR #60, opened 2015-05-29
pass event object to click callback

RacingTadpole:
I'd like to have click behaviour depend on whether the shift key etc is pressed - so would like the event object to be passed along to the click handler callback.  What do you think?


dkoes:
sounds good to me

dkoes:
I made the same change (6b54386) just also pass the container

PR #64, opened 2015-06-05
check sel and style are valid in model.setStyle

RacingTadpole:
I've just spent an hour or so trying to get setStyle to do what I wanted, and just discovered that I was typing "hetatm: true" instead of "hetflag: true".  I realised what was happening when I accidentally typed "spheres" instead of "sphere", and saw the atoms disappear, just like "hetatm" was doing. 

So, to save myself and others this time in future, I've made this change so that 3Dmol logs to the console if it receives an unknown selector or style.

Or you might have a better idea than just logging to console - e.g. should it throw an error?

Thanks!

### DIFF
--- a/3Dmol/glmodel.js
+++ b/3Dmol/glmodel.js
@@ -52,6 +52,51 @@ $3Dmol.GLModel = (function() {
         "Ni" : 1.63
     };
 
+    var validAtomSpecs = [
+        "resn", // Parent residue name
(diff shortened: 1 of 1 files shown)


dkoes:
I think it's better to avoid throwing errors.  I wonder if there's a way to more tightly integrate this with the JSDoc..

PR #68, opened 2015-06-09
Add public function updateContainer(element) to GLViewer

RacingTadpole:
I am using 3Dmol on a page whose DOM can change if the user clicks on an "about" or other link. When the user comes back to the molecular view, I need it to be able to show the same GLViewer again, in a new container element. (An alternative would be to just hide and re-show the container, but I'm using ReactJS to handle the view changes, and it would be hard to change its behavior.)

So, I have put some of the GLViewer constructor into a private function initContainer(element), which the constructor calls.  So far, no change in the input/output (the git change log looks long, but that's mostly because I've indented lots of lines, since they're now inside another function).

The important interface change is: there is also now a public function updateContainer(element), which reconnects the canvas and mouse clicks to the new container element.

This is working quite nicely for me.  To see it in action, you can literally use it to move the GLViewer from one DOM element to another - eg. on the tests/example.html, change the `gldiv` line to:

    <div id="gldiv" style="width: 50%; height: 80vh; margin: 0; padding: 0; border: 0; float:left"></div>
    <div id="newdiv" style="width: 50%; height: 80vh; margin: 0; padding: 0; border: 0; float:left"></div>
    <div style="clear:both"></div>

And then in the console, switch the molecule back and forth between the divs with:

    glviewer.updateContainer('newdiv')

and

    glviewer.updateContainer('gldiv')

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -282,12 +282,8 @@ $3Dmol.GLViewer = (function() {
             HEIGHT = container.height();
             ASPECT = WIDTH / HEIGHT;
             renderer.setSize(WIDTH, HEIGHT);
-            console.log('renderer', renderer);
-            console.log('container', container, container.children());
             container.append(renderer.domElement);
-            console.log('post-append container', container, container.children());
             glDOM = $(renderer.domElement);
-            console.log('glDOM', glDOM);
 
             if (!nomouse) {
                 // user can request that the mouse handlers not be installed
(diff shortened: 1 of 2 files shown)


dkoes:
Would changeContainer perhaps be more descriptive?

dkoes:
I changed it to setContainer

RacingTadpole:
Perfect, thanks.

PR #80, opened 2015-06-22
Unit Cell

jlc206:
Capability to add unit cell shape to PDB models


dkoes:
fe69130

The GLViewer class now has an addUnitCell function.  If a model has unit cell data (e.g., from a pdb file) this will return a GLShape object with the box defined as line shapes.

PR #84, opened 2015-06-22
add support for ChemDoodle JSON

jus1in:


dkoes:
next step.. test with openbabel conversion

PR #87, opened 2015-06-25
documentation completed & addStyle wireframe opacity bugs fixed

Jenny19880324:


dkoes:
This needs to be cleaned up a little bit.  glmodel.js~ is a temporary file that should not be added to the repository.  glshape.js has a random url embedded in it (http://www.w3schools.com/jsref/dom_obj_canvas.asp).  The indentation is not standard - should be all spaces.  

I'm looking forward to an improved test suite though.

Jenny19880324:
I removed the glmodel.js~ and the random url in glshape.js.
What do you mean by indentation should be all spaces? Do you mean I have tabs as indentation?

dkoes:
If you look at the diff of render.js, the indentation is inconsistent, which I assume is due to mixing tabs and spaces.

PR #89, opened 2015-06-27
Fix most of the problems with mmCIF & CIF

jus1in:
This fixes most of the issues with mmCIF and CIF. It now correctly parses all of the sample files. However, I noticed some issues with some of the files. It does not seem to correctly infer all of the bonds. 
Some examples:
 * [PANCUC](http://jus1in.com/3Dmol.js/viewer.html?=/testFiles/PANCUC.cif&style=stick;cross)
 * [JARMEU](http://jus1in.com/3Dmol.js/viewer.html?=/testFiles/JARMEU.cif&style=stick;cross)
 * [NARLEY](http://jus1in.com/3Dmol.js/viewer.html?=/testFiles/NARLEY.cif&style=stick;cross)

It seems to happen more often than not.

### DIFF
--- a/3Dmol/parsers.js
+++ b/3Dmol/parsers.js
@@ -492,16 +492,6 @@ $3Dmol.Parsers = (function() {
             return sections;
         }
 
-        // Parser puts all of the data in the file in an object
-        // uses getDataItem() to get an array for the data item
-        // given. The data item has the category name prepended to it.
-        // The possible categories and data items in each category are defined
-        // in
-        // the mmCIF specification
-        function getDataItem(dataItemName) {
-            mmCIF[dataItemName] = mmCIF[dataItemName] || [];
-            return mmCIF[dataItemName];
-        }
 
         var lines = str.split("\n");
         // Filter text to remove comments, trailing spaces, and empty lines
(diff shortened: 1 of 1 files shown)


jus1in:
It looks like this is not a new issue. I Looked at files that "worked" before and they have this same problem before and after these changes. [Example](http://jus1in.com/3Dmol.js/viewer.html?=/testFiles/5HVP.cif&style=stick;cross)

dkoes:
I just looked at PANCUC.  This does not have any bond information in it so we must compute the bonds.  We are missing a few, probably because we need larger bond distances for elements like gold and antimony.

We do not pick up on the triple bond nature of the N-Au bond, but neither do pymol or jmol when given the cif files (the mol2 files from CCDC has this bond explicitly in it).

The problem with the bond detection can be easily fixed.  See issue #90.

dkoes:
I measure the distance between the N and C in NARLEY to be 1.47417 (using pymol) so it should be picked up as a  bond.

**Incidentally, the 'atom' field of the atom object should be set to the atom name (_atom_site_label).** (I'm bolding this to remind myself to check that this side-issue is fixed before this issue is closed).

Stepping through the code, it looks like the N-C distance is a little longer than pymol thinks it is.

I have some doubts about the calculation of the x,y,z coordinates since the unit cell in this case does not have all 90 degree angles.  See [this wikipedia article on fractional coordinates](https://en.wikipedia.org/wiki/Fractional_coordinates) that has been tagged as possibly having incorrect math in it.

jus1in:
fixed. d9979852de16b7030c696ff09a3f326205062873
The math on the wikipedia page was correct.

dkoes:
Fantastic!

PR #114, opened 2015-07-30
test cases of setViewStyle added

Jenny19880324:
only commited tests foler


dkoes:
This is merged, but the previous test image for test24 looked a lot better than the current image...

Jenny19880324:
Hi David,
This is because I make the z-offset of the outline shader much smaller now.
If the z-offset is larger, the outline is generally better for all geomtries(both the surface and the sticks), but it produces the artifact that the outline would be gone within a zooming range. I think it might be because the z-off is too large so that it goes out of the clipping range.
(I tried turning off the lambertdouble shader and only turning on the outline shader. It turned out that both the offset outline and the original part would be gone within a certain zooming range. That's why I am saying it's probably being cut off by clipping. But I am not sure why it's coming back after zooming further out. )
I don't think I can solve this because it's really difficult to predict the thickness or curvature of the geometry that is to be rendered.

Maybe we can say for cartoons, use a smaller z-offset and for surface and sticks, use a large z-offset. But it's hard to say because you can't really know how the surface and sticks look like.

PR #119, opened 2015-08-12
refactor processSymmetries to not care about format

jus1in:
Refactored processSymmetries to not care about format to avoid code duplication and make using using this method in other formats easier. I am not merging yet because I am not sure if there was a reason @jlc206 used the approach she did, and the more general approach might break things I haven't found.


dkoes:
@jlc206 please review this change.

PR #124, opened 2015-09-03
Multimodel support/Animate func

jlc206:
Support for multimodel files - 
This was done by altering the parser so that depending on the options sent to it, it will either read all models into lists, just read the first model, or read all atoms from all models into a single list. The default is still to only read the first model. The parser will now return either a list of atoms, a list of lists of atoms, or an empty list depending on the option it is sent. 
Also added 3 functions to glviewer: addModels, addModelsAsFrames, and addAsOneMolecule. addModels adds all of the models in a multimodel file to the viewer as separate models, addModelsAsFrames adds just one model to the viewer and fills a model property called "frames" with the lists of atoms for other models and addAsOneMolecule adds all of the models' atoms in a single model.

Animate -
A viewer with at least one model that has been added using the "addModelsAsFrames" function can then be animated. Animate takes 3 optional parameters of the interval (speed), loop direction (forward, backward, or backAndForth) and number of repetitions.

### DIFF
--- a/3Dmol/glmodel.js
+++ b/3Dmol/glmodel.js
@@ -111,6 +111,7 @@ $3Dmol.GLModel = (function() {
     function GLModel(mid, defaultcolors) {
         // private variables
         var atoms = [];
+        var frames = [];
         var id = mid;
         var hidden = false;
         var molObj = null;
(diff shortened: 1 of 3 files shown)


jus1in:
If I am understanding your implementation correctly, each parser first parses its file into models and their atoms. It then puts all of the atoms into the `atoms` array. It also returns an array of models, which are arrays of atoms. The two different places where each atom object is used creates a problem.

Currently, atom bonds work by specifying the index of the other atom in the array. With just a single model, this means that each atom stored a list of indices from the `atoms` array. These same indices will not, however, match up with the indices of the atoms when put into the array of models. To fix this each atom object would have to be cloned and then all of the indices changed in the clone. This is not a very good solution.

I propose removing the atoms array. The parsers will only return an array of arrays. If a caller, like `addAsOneMolecule`, needs a single array with all of the atoms, it can concatenate the arrays and change the indices on its own, probably with a helper function.

dkoes:
I agree there isn't any need to provide an atoms argument to parse any more as this information is redundant to the return value.

It isn't clear to me from glancing at the code that all the parsers are doing the right thing nor that the case of addModelData being called multiple times would work properly.  

I recommend making some test cases for these situations.

Issue #11, opened 2015-01-14
support for multiple cartoon formats


dkoes:
edged (current), oval (jmol/pymol), flat, basic (cylinders for helices)
also thickness and height
nucliec sticks

mattragoza:
Do we still want the option to display alpha helices as simple cylinders?

dkoes:
yes, and a ribbon option

mattragoza:
Do you have an example of what the ribbon style should look like?

dkoes:
Basically don't do anything special for helices and sheets (all loop style).  If we could have both a tubular style and a flat style like that it would be great.

mattragoza:
I've added a ribbon option that forces the width of the strand to follow the thickness property rather than being secondary-structure-dependent. This can be specified in addition to the style property, allowing ribbons in oval (tubular), default (edged), rectangle ("rounded edged"), etc. The thickness of any strand can be set to 0 for a "flat" style. For alpha helix cylinders, I'm currently using $3Dmol.GLDraw.drawCylinder(), which has an option for either rounded caps or no caps at all- is this ok for now or should I modify the function to allow flat caps as well, which appears to be what pymol uses for basic alpha helices?

dkoes:
I think we have enough styles now.

PR #126, opened 2015-09-19
beta sheet arrows, alpha helix tubes, ribbon style, and tests

mattragoza:
Beta sheet arrows can now be specified by the 'arrows' flag of cartoonStyleSpec, alpha helices can be drawn as simple cylinders by the 'tube' flag, and strands can be depicted as a constant-width ribbon using the 'ribbon' flag- as prescribed in #11.

In addition, click detection accuracy in new strand styles (oval, rectangle etc) has been improved, 'opacity' can now be implemented in flat strands ('thickness' set to 0), and various additional bug fixes related to models rendered with multiple cartoon styles.

A setClickable() function has been added to GLModel and GLViewer, allowing a more user-friendly interface for applying clickable functionality and callback functions to atomSelectionSpec objects, similarly to how setStyle() is implemented.

Additional test pages have been added to check.html reflecting these updates.


dkoes:
There are still a few issues when applying a cartoon style to a molecule that doesn't have secondary structure.  Go through the files in tests/test_structs that aren't proteins (not pdb).  Applying a cartoon style to something that isn't a protein/nucleic acid should show nothing and it should not generate any errors.

I'm getting
    `TypeError: curr is undefined`
with benzene.mol2.

AllanSCosta:
Hi! 

I believe there's currently no way to display a cartoon with an alpha carbon-only pdb. Is there a way around it? Based on this request it seems that 3dmol checks for a fully specified pdb file? 

Thanks!

PR #128, opened 2015-09-22
improved multimodel support/tests

jlc206:
added multimodel support for sdf, xyz, mol2, and pqr file formats. @jus1in said he'd be willing to add multimodel support for mmcif files so after that we should have all the parsers working for multimodel file support (excluding cube & json - which you said we shouldn't worry about). The default behavior is still to read just one model in, but if the options "multimodel" or "onemol" are sent in, all the models in the file will be read. Setting these options is handled by whichever viewer function is called when adding the model (addModelAsFrames, addAsOneMolecule, etc.)

On the first call of addMolData for a new model: if only one list of atoms is returned from the parser, this list is added to the model's frames and the model's atoms is set to this. If a list of atom lists (frames) is returned from the parser, these lists are added to the model's frames and the model's atoms are set to the first frame

On subsequent calls to addMolData: atoms will be added to the current frame, unless the option "frames:true" is sent - then atoms will be added as new frames


dkoes:
Has the documentation been updated?

PR #131, opened 2015-09-30
Store cell info from Cif files

### LINKED ISSUE #130: CIF parser doesn't store cell info
The cif parser parses the unit cell parameters, but doesn't store them in the model.cryst variable as the PDB parser does.

sbliven:
Fix #130


sbliven:
This section seems to have a lot of unnecessary variables that aren't ever used, but I took a conservative approach and left them in there.

Issue #132, opened 2015-09-30
Enhance animation API: stopping an animation, querying if molecule is being animated


cryos:
We have been using the new animation API and would like a way to stop the molecule animation, and also some API to query whether the molecule is currently being animated or not. As far as I can see these are both missing right now.

cryos:
This has been fixed in commit 65dab03.

PR #134, opened 2015-10-02
fix issue #132

jlc206:
Added glviewer function "stopAnimate" which stops animation of a viewer, and "isAnimated" which returns true if a viewer is currently being animated, and false if the viewer is not

Replaced test 38 with a very similar test that displays usage of these new features

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -1358,6 +1358,38 @@ $3Dmol.GLViewer = (function() {
 
             return s;
         };
+        
+        /**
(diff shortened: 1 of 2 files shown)


dkoes:
Can we have setFrame and getFrame methods in the viewer class?  Then refactor animate to use these methods (less overall code).

jlc206:
let me know if this is what you mean. The setFrame just sets all the models in the viewer to whatever frame (very similar behavior to the glmodel's setframe) but the getFrame just returns the highest number of frames that a model in the viewer has - this is different from the glmodel's getFrame (which returns the frames list)

PR #142, opened 2015-10-23
Fix modelData when adding multiple models at once

jus1in:
This fix touch a *lot* of code, so I hope it doesn't break anything. I did my best to test everything, but I might have missed something.

Edit: just realized this also includes all of the code to make mmCIF parse multiple molecules.


dkoes:
What's GLModel.setAtomDefaults?

PR #144, opened 2015-11-10
added custom PDB URI option for the download function

onursumer:
This allows users to customize previously hard-coded PDB download URL. This is especially useful when requesting PDB data through a proxy within a secure (HTTPS) web application.


dkoes:
Looks good to me, any chance of adding a JSDoc comment?

onursumer:
I am not sure where to add the comment. One option is to extend the definition of the <code>@param {Object} options</code> to multiple lines and explicitly document each option. I think for now we only have <code>options.format</code> and <code>options.pdbUri</code>.

dkoes:
Yup, that's where I'd put it.

PR #168, opened 2016-04-13
Added vasp POSCAR/CONTCAR files model parser

alejandrogallo:
This parser needs that the atomic symbols are explicitly provided in 
the POSCAR file, this is in VASP optional but for parsing purposes 
necessary. If this were not the case a warning is echoed.


dkoes:
Thanks!

alejandrogallo:
Hi ! No problem, I have done a parser for vasp CHGCAR volumetric data, I'm creating a pull request, I also added a test with sample data.

PR #169, opened 2016-04-14
Vasp parsers

alejandrogallo:
Added volumetric data for CHGCAR vasp files.
TODO: Also for PARCHAR AND CHG, I have 
added a source link where this could be implemented from.


dkoes:
I think other codes (e.g. PyMOL) interpolate the grid values whereas we fully discretize the voxels.  Smoothing is an approximate way to get some interpolation, but we could do better.

This merits its own issue: #171

alejandrogallo:
It is correct, you are right it does seem offset but it is like that because of the "isovalue value" 
if you take a higher one like 0.042 then you can see the high charge density at the bonding regions.

dkoes:
Okay, I've added test46 to check.html that incorporates this structure.

PR #182, opened 2016-05-27
Providing support to AMBER topological files (prmtop file format)

Manohar-94:
1. Created a parsers.prmtop function inside the parsers.js file for parsing the prmtop file. 
2. Created a glmodel.setCoordinates function for updating the coordinate data from the inpcrd file. 
3. Made a test case test50.html.


dkoes:
Looks good.  The next step is to support full trajectories in setCoordinates (e.g. mdcrd).

PR #190, opened 2016-06-25
Provide support to gro file format

Manohar-94:
1. Added a parsers.gro function to parse gro files.
2. Created a test case test54.html


dkoes:
Great!

PR #193, opened 2016-07-10
Load ANISOU values from PDB file

dllahr:
This code extends the loading of PDB files to also load the values from the ANISOU line/field.  It loads it into the property "uMat" on the atom.  Assumes valid PDB format e.g. ATOM line occurs before ANISOU, etc.


dkoes:
Thanks!

Issue #112, opened 2015-07-24
more robust colorschemes 


dkoes:
should support user defined functions and have built in color by residue, structure, etc

dkoes:
We have colorfun now, but it isn't documented (also an issue raised in #194)
Should have builtin color by chain.
The easiest way to do this is to define property maps that can then be provided to colorscheme

dkoes:
Here's one possibly mapping for chains:
https://gist.github.com/carlosmr12/dca1caa6973b9e1e784cca9507a23017

helblazer811:
So you want people to be able to pass a dictionary with all of the colors for each chain. That does not sound incredibly difficult.

PR #197, opened 2016-07-15
Built in Colorschemes 

helblazer811:
This resolves issue #112 by allowing users to pass a string to the colorscheme property which takes from the builtinColorSchemes map. See colors.js.


helblazer811:
I checked all of the tests and the ones that are affected by this work fine as well as the ones that originally used colorschemes.

PR #199, opened 2016-07-21
PDB file parsing by the setCoordinates function

Manohar-94:


dkoes:
I would do a shallow atom copy regardless of the format type to be safe (e.g., if the topology was loaded using pdb instead of top). Also, the merge conflicts need to be fixed.

PR #201, opened 2016-07-26
This has the resolution to the secondary colorscheme issue and also some work on viewer.html that is incomplete

helblazer811:


helblazer811:
The secondary colorschemes are resolved. Issue #204 should be resolved. I just removed the zoom functionality from zoomTo and put that into a center function. Viewer.html should work shortly; i currently have the parser working so that an object is created representing the url. The user will be able to easily manipulate this url. There are also image_tests and revised_testing which are still in development

PR #203, opened 2016-07-28
Explicit Typed Array Length

justinmc:
Typed array constructors should throw an error when called with no arguments according to the [ecmascript spec](http://www.ecma-international.org/ecma-262/6.0/#table-49).  There are a few places in this project that don't pass arguments, so this PR fixes them by just passing `0`.

The latest version of Chrome doesn't throw an error like it should, so most people probably don't notice that this is wrong.  However, [core-js](https://github.com/zloirock/core-js), which is used in Babel's popular [polyfill](https://babeljs.io/docs/usage/polyfill/), does throw an error and causes 3Dmol to break entirely.

Let me know if there are any other places in the codebase I missed!


dkoes:
Thanks!

Issue #205, opened 2016-07-29
atom.color vs atom.style.color  


sillitoe:
Took me a while to figure out what was going on with this...

I create a new viewer:

```
let opts = { defaultcolors: $3Dmol.rasmolElementColors };
let v = $3Dmol.createViewer( $('#target'),  opts );
v.addModel( pdbdata, "pdb" );
v.zoomTo();
v.render();
```

then colo(u)r it via:

```
v.setStyle( {}, { cartoon: { color: 'red' } } );
```

later on I try and change the colour (according to a ```conservation``` property set elsewhere):

```
v.getModel().setColorByFunction({}, function(atom) {
        let c = myColorScale( atom.properties.conservation );
        return c;
 });
v.render();
```

However, this didn't make any difference to the coloring.

Eventually figured out that this was due to:

* ```setColorByFunction()``` changes ```atom.color```
* ```setStyle``` changes ```atom.style.color``` 

(the latter takes preference).

https://github.com/3dmol/3Dmol.js/blob/master/3Dmol/colors.js#L451-L454

dkoes:
Yeah, I want to deprecate the atom.color and setColorByFunction methods.  Instead you can provide a colorfunc as part of the style (also property map colorschemes).  This is not properly documented.

sillitoe:
No problem - figured it might be better to choose one or the other. Providing a colorfunc function sounds ideal.

Happy to contribute - could try and put together another tutorial page? The docs are already good, but there were a few gaps where I needed to drive into code.

dkoes:
Pull requests with better documentation/tutorials would be very welcome.

The long range plan is to have a test system that automatically extracts test cases from examples embedded in the documentation, but that's a long way off..

sillitoe:
Sounds like a plan.

So the usage you want to encourage is something like...

let colorAsSnake = function(atom) {
  return atom.resi % 2 ? 'white': 'green'
};

viewer.setStyle( sel, { cartoon: colorfunc: colorAsSnake  });

?

(sorry for typos - on a train)

dkoes:
yup

dkoes:
#214 provides improved documentation.

PR #208, opened 2016-08-01
first commit for the api tutorial

sillitoe:
Covers a basic introduction to some of the aspects of the API that would've avoided a couple of gotchas (#205)

Would definitely be nice to separate the code into testable sections, but works fine on my local build.

Note: I've included the PDB file in a new directory ```./doc-data/``` (which might not be ideal).


dkoes:
I like the try it buttons

PR #209, opened 2016-08-01
fix minor inconsistency between example and code

sillitoe:
just noticed these

would be great to be able to show 'live' code to avoid copy/paste problems


dkoes:
Easy enough to do. I've pushed an implementation (91ff482)

PR #212, opened 2016-08-02
Fix missing variable declaration

justinmc:
A variable was used without a declaration, which is either going to automatically declare it on `window` or throw an error, both of which are bad!  This fixes it by declaring the variable.


dkoes:
Thanks! At one point we had automated checks for this sort of thing..

PR #215, opened 2016-08-09
Declare undeclared cylinder vars

justinmc:
One more batch of undeclared variables that started popping up in my console.


dkoes:
Thanks!

Issue #217, opened 2016-08-17
document the options paramter to addModel


dkoes:
The options supported differ based on the file format, but we don't even mention it now.

helblazer811:
I think we should create a specification for these options because it is used in multiple function.

helblazer811:
We definitely need to define the options for the different parsers somewhere.

dkoes:
http://localhost/3dmol/doc/$3Dmol.html#download  is just one place we reference options without defining them.

helblazer811:
What I have done so far is construct a list of all of the parameters that are used for those functions and i am calling it ModelSpec.

helblazer811:
/**
  * Model specification. Used to specify the options of a GLModel.  Depending on the input file format, not all fields may be defined.
  * @typedef ModelSpec
  * @prop {boolean} duplicateAssemblyAtoms- Set to true if you wish to diplicate assembly atoms otherwise false.
  * @prop {boolean} frames - true if you want to add to a new frame and false otherwise
  * @prop {object} vibrate - object specifying the vibration behavior
  * @prop {number} vibrate.frames - number of frames to be created, default to 10
  * @prop {number} vibrate.amplitude -amplitude of distortion, default to 1 (full)
  * @prop {boolean} multimodel - specifies weather or not multiple models are being defined .supported by xyz,sdf, or mol2
  * @prop {boolean} onemol -specifies weather or not the model is of one molecule . Supported by xyz , sdf , mol2
  * @prop {boolean} keepH - specifies weather or not the porgram should keep H . supported by sdf,mol2
  * @prop parseStyle - cdjson
  * @prop {boolean} doAssembly - mcif
  */

//this is an unfinished attempt, i need to make sure no other things are used

dkoes:
These are properties of the model, but options for the parser.  I would call it ParserOptionsSpec, and not all are implemented on all file types.

helblazer811:
Ok. I should be able to do that, and I will also go back and change the
type of option in all of the functions it appears in. I had noticed that
the navigation bar on the website animates down based on the top of the
page not the bottom of the Nav bar. I created an issue for it.

Issue #196, opened 2016-07-14
Improved Test Architecture


helblazer811:
Create a single page that displays all of the tests. Make something that dynamically loads in all of the examples into the documentation with a display of how it looks. The test page should have a fixed canvas on one side of the screen and images of the different tests on the right side. The tests on the right  side can be clicked and that displays them in the canvas on the left side.

helblazer811:
I have created a page that loads all of the reference images from the tests,then loads all of the javascript from those pages then takes a picture of the canvas and displays that. So you have two columns of images. It works for the most part except for surfaces which load after the model is rendered.

dkoes:
Should prioritize the test for non-local variable declarations

helblazer811:
Could you elaborate?

dkoes:
Every test case should be wrapped with code that makes sure no variables (other than $3Dmol) should be added to the global namespace.  You can check this by recording all the variables in window before calling the code and making sure there's nothing new after calling the code.

See saveGlobal and checkPollution in tests/qunit.js

helblazer811:
Yeah the viewer variable is not necessary I decided to do thst half way in.
I have just yet to fix some of the test cases.

dkoes:
There's still more to be done here, yes?
 * autogeneration of examples from documentation
 * auto comparison of images
 * accurate timing of surfaces
 * automated running of tests and recording of trends, associate any regressions with code
 * global value pollution checking? error checking?

helblazer811:
Ok. I guess this issue is very broad. I was thinking about the image generation.

PR #222, opened 2016-08-22
Image Tests and Parser Options

helblazer811:
This resolves issue #217 and resolves issue #196. Note that the 70 files include all of the tests so that they work with the new testing architecture. The image tests use an arbitrary amount of time for the tests with surfaces because they are loaded asynchronously ; that needs to be fixed.


helblazer811:
This also includes work on the viewer.

PR #226, opened 2016-08-26
This is the working prototype of the image tests

helblazer811:
There is still more work to be done, but it can be used. The time stamp for the surfaces are not currently an accurate depiction of the amount of time that it takes to render. Resolves issue #196.


dkoes:
Can you demo it for me on Monday?

helblazer811:
Sure. Are we still meeting with avagadro?

Issue #229, opened 2016-08-28
let addSurface take string for surface type and generate log messages if an invalid type is passed


dkoes:
In the viewer object, let's let people pass "VDW" instead of $3Dmol.SurfaceType.VDW.  Also, if an invalid type is passed, we should print out a log message (and the default should probably be VDW, not SAS).

helblazer811:
Should i do something similar to what i did with colorschemes and create a map?

helblazer811:
It should work now. I created a pull request.

PR #230, opened 2016-08-28
resolves issue #229

helblazer811:
Both this: 
viewer.addSurface("VDW", {opacity:0.85, voldata: data, volscheme: new $3Dmol.Gradient.RWB(-10,10)},{chain:'A'},callback);

and this work:
viewer.addSurface($3Dmol.SurfaceType.VDW, {opacity:0.85, voldata: data, volscheme: new $3Dmol.Gradient.RWB(-10,10)},{chain:'A'},callback);
See test # 19.


dkoes:
Should also print a message if passed a non-string that isn't a built-in surface type.

helblazer811:
Should do that now.

Issue #139, opened 2015-10-16
support boolean selector fields


dkoes:
Have  `and` and `or` fields of a atom selector object that contain a subquery to combine with the current.

Or, less ugly - take lists of subqueries that are combined and require that selection objects with an `and` or `or` can contain no other fields (except model).

helblazer811:
So this is essentially set intersection?

dkoes:
Yes, or union for or.

Issue #164, opened 2016-04-09
strange double bond orientation


dkoes:
http://3dmol.csb.pitt.edu/viewer.html?cid=307900&style=stick:radius~0.2

PR #245, opened 2016-11-14
Boolean selector fields

helblazer811:
This resolves issue #139 and i added a test for issue #164. See test #68-70 for the boolean format.


helblazer811:
I also added not for closure's sake.

dkoes:
We had invert...

PR #249, opened 2016-12-11
fix tests with PhantomJS

lfrohman:
Our company uses 3Dmol and we run front end selenium tests using PhantomJS as part
of our build process. With PhantomJS, window.URL is null, so we need to use this simple test.


helblazer811:
Ok thanks. Feel free to submit any more pull requests or issues on features that you may enjoy.

lfrohman:
thanks, I have one more change, I will work on it today.
Also, we use custom attributes for the atoms, we added our custom
attributes to the array

validAtomSelectionSpecs

so that we don't see the console message "'Unknown selector  xxx"
but I am sure this is not a good way to do this.
Should there be a
function addValidAtomSelectionSpecs(customSpecs)
?
Or do you have a good way to do this?

thank you,
Lance Frohman

dkoes:
Sounds reasonable to me.

PR #251, opened 2016-12-12
customize hover duration

lfrohman:
Currently the hover functionality is triggered after 500 milliseconds
This change allows the client to customize the amount of hover time before the hover
functionality is triggered.


dkoes:
Looks good. Might be a little better if the hoverDuration could also be set via a config option of the viewer at construction time.

PR #253, opened 2016-12-12
function addAtomSpecs

lfrohman:
this allows custom atom specs to prevent the 'Unknown Selector x' message on the console for the strings passed


dkoes:
Thanks

PR #264, opened 2017-01-23
Fix bug with default background color being white

danielholmes:
This change https://github.com/3dmol/3Dmol.js/commit/dd8674a6a8f0f5a65091149ffad70f0cdb7ff574#diff-530cc22d3d23d878b3619ab210a374a3R30 effectively hardcoded the background color to white since it overrides any value someone would pass in as config


helblazer811:
Thanks! If you have any other issues that you would like us to look at let us know.

Issue #295, opened 2017-04-16
Question regarding VASP support + crystallographic data generally


mkhorton:
Hello,

I hope this is an appropriate place to ask.

I can see that 3Dmol's volume class has a 'vasp' option, and there are some references to POSCAR files in the code too, but I can't find documentation for this anywhere. Does 3Dmol.js support visualization of VASP POSCAR/CHGCAR files? (Edit: I've now found the vasp_test.html file in your source, so will try this out!)

More generally, how robust is 3Dmol.js' support for crystallographic data generally? i.e. is it possible to draw the unit cell bounding box, axes compass, or co-ordination polyhedra (if bonds have been detected)?

Thank you kindly for your time!

mkhorton:
As a follow-up, these CHGCAR files can be very large (10s-100s MB) and not suitable for providing directly online. If 3Dmol can derive an isosurface, is there a way to serialize this isosurface, so that we can do the generation once and then supply the user with the derived isosurface instead of the CGHCAR, to save time loading/processing?

dkoes:
You can call addUnitCell to draw the unit cell bounding box, but this currently only works with pdb and mmcif file formats.  We could certainly add support for other file formats if they have the required information.

Can you give examples of axes compass and co-ordination polyhedra?

You can use addMesh to provide pre-rendered isosurfaces.

mkhorton:
Sure,

Axes compass: this is simply three arrows showing the direction of a, b and c lattice vectors defining the unit cell bounding box, e.g. see the arrows in this [image](https://www.researchgate.net/profile/O_Parasyuk/publication/258922075/figure/fig1/AS:297314498433024@1447896708322/Figure-1-A-unit-cell-of-Cu-2-CdGeS-4-Drawn-by-VESTA-40.png)

Co-ordination polyhedra: instead of a ball+stick atoms+bonds representation, it's common to draw spheres for atoms, and then join all the atoms an atom is bonded to together to form a polyhedron, example: https://www.materialsproject.org/materials/mp-12992/ This is a very common approach used by e.g. JMol, Vesta, CrystalMaker etc.

Regarding the vasp_test.html file, I've got this working on my computer (macOS, Safari, latest versions of each), but the web inspector complained about a missing 'chgcar.list' file the test page failed to load, I had to copy some files around to get it to work. I'm not sure if this was error on my part or a bug in the test file.

> You can use addMesh to provide pre-rendered isosurfaces.

Thanks, will look into this.

[...]

PR #296, opened 2017-04-18
Update VASP parsing for `addUnitCell` to work for POSCAR, CONTCAR files

mkhorton:
Referencing #295 

My JavaScript is a bit rusty, but I think this should work? Would be happy to contribute further patches once my understanding of 3Dmol.js is improved.

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -2005,21 +2005,29 @@ $3Dmol.GLViewer = (function() {
             s.shapePosition = shapes.length;
             var data = model.getCrystData();
             if (data) {
-                var a = data.a, b = data.b, c = data.c, alpha = data.alpha, beta = data.beta, gamma = data.gamma;
-                alpha = alpha * Math.PI/180.0;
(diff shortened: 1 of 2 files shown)


dkoes:
Looks reasonable to me.. any chance of getting a test case added to tests/auto/tests?

mkhorton:
Sure, will look into it.

mkhorton:
Sorry, I'm a little confused as to how the auto tests work. What's the purpose of the cgi script, can these tests only run when deployed on a server and not locally?

I tried opening tests/index.html, but clicking the static image just links to tests/auto/generate_test.cgi?test=test1.

I think a good place to test this would be in test46, I've added a change here:

https://github.com/mkhorton/3Dmol.js/commit/7f30db85e30e594af5bf3ade9a831d4f315ac165

However I'm not sure how to actually deploy this to see if it works? Many thanks.

dkoes:
Yes, to view the tests individually you need to be using a webserver, but you still should be able to load them all: http://3dmol.csb.pitt.edu/tests/auto/tests.html

PR #302, opened 2017-05-21
Added support to lammpstrj file format

Manohar-94:
Created a parsers.lammpstrj function to parse lammps trajectory file.
Added test case - test90.js


3Dmol/parsers.js
```diff
@@ -2202,5 +2202,52 @@ $3Dmol.Parsers = (function() {
...
+                        else
+                            atom[prop] = parseFloat(tokens[j]);
+                     }
+                     atom.bonds = [];
+                     atom.bondOrder = [];
+                     atom.properties = {};
+                     atoms[atoms.length-1][i-offset] = atom;   
+                }
+            }
+            lines.splice(0,offset+atomCount-1);
```
dkoes: splicing isn't particularly efficient as it has to construct a new array.  You should avoid modifying the array.  You can have a nested loop structure where the inner loop updates the iteration index of the outer loop (pseudocode):
```
for(var i = 0, n = lines.length; i < n; i++ ) {
 if (lines[i].match(/ITEM: NUMBER OF ATOMS/))
                   atomCount = parseInt(lines[i+1]);
else if (lines[i].match(/ITEM: ATOMS/)) {
   for(var stop = i + atomCount; i < stop; i++) {
   ...
  }
}
```

3Dmol/parsers.js
```diff
@@ -2202,5 +2202,52 @@ $3Dmol.Parsers = (function() {
 	}
         return atoms;
     }
+
+    /**
+     * Parse a lammps trajectory file from str and create atoms
+     */
+    parsers.lammpstrj = parsers.LAMMPSTRJ = function(str, options){
```
dkoes: Isn't this a trajectory file (ie, it needs a topology file to be complete)?  If so it should be in GLModel.setCoordinates, right?

Manohar-94: In lammps, we dont have the concept of a topology file. All the information is present in the trajectory file itself.

3Dmol/parsers.js
Manohar-94: Done

3Dmol/parsers.js
```diff
@@ -2208,45 +2208,44 @@ $3Dmol.Parsers = (function() {
...
-                break;
-            var offset = 0;
-            var atomCount = 0;
-            for (var i=0; i<lines.length; i++){
-                if (lines[i].match(/ITEM: ATOMS/)){
-                    offset = i+1;
+        var offset = 0;
+        var atomCount = 0;
+        var start = 0;
+        while (start<lines.length-9){
```
dkoes: I don't understand this -9 check.  What if you have a small system with only 3 atoms? You would miss the last frame.

Manohar-94: Even if there is 1 atom, the file would have 10 lines. This just provides cushion, for example, in case the last line is empty.

dkoes: Are you sure the box lines are always there? What if periodic boundary conditions weren't used?  Seems like it would be more robust to just absorb the lines rather than assume a fixed sized header.

Manohar-94: The box lines are always there regardless of whether the boundary conditions are fixed or periodic or shrink wrapped.

PR #307, opened 2017-06-13
Changes to Viewer

helblazer811:


helblazer811:
Note that centering is only fixed when you click the center button on the side panel. It does not center correctly to begin with for some reason.

dkoes:
Still not using specStringToObject?

dkoes:
Clicking on the hamburger button doesn't work in an empty viewer.  This line is erroring out.
`  currentSelection.dumps.push(tokens[token]);`

PR #308, opened 2017-06-18
support added to netcdf trajectory file

Manohar-94:


dkoes:
Your test case is adding a variable `reader` to the global namespace.  This needs to be fixed.  This was caught during the automated tests.

PR #313, opened 2017-06-21
All updating works 

helblazer811:
This should work as a first prototype. All updating works upon pressing render. You can add attributes and change values, but you can not delete things currently and it is buggy.


dkoes:
Let's at least get to the point where there are no known bugs to merge..

helblazer811:
The only real bugs that i have observed occur when the user renames a style to have the same selection as another style resulting in conflict.

dkoes:
if they are being applied in the order specified by the user, I'm not sure that's a bug

helblazer811:
whats happening is the url contains two selections chain:A and chain:B , both contain one style selection, and the user is changing chain:B to chain:A so there are now two styles for the same selection.

dkoes:
In that case the second should override the first.

Issue #315, opened 2017-06-22
support atom selections in place of coordinates


dkoes:
For example, the start/end of a line specification could be atom selections. The center of the selected atoms would be used.  Basically, if the start/end objects don't have x/y/z fields but are objects, assume they are selections

helblazer811:
I could do it by just taking the selection and replacing it with the coordinates as right after the user passes it. This will mean that no changes to the code other then this conversion need to be made.

dkoes:
but you need to verify that it isn't an object that has x,y,z members.  Also, you want to take the average of all the selected atoms.  This would be in a function that could be called by all the methods in draw that have start/ends

PR #320, opened 2017-06-29
added function getUniqueValues and changed test 92 to use resi select…

helblazer811:
…ions


helblazer811:
closes issue #315

3Dmol/glviewer.js
```diff
@@ -1055,6 +1055,31 @@ $3Dmol.GLViewer = (function() {
...
+        * @param {string} attribute
+        * @param {AtomSelectionSpec} sel
+        * @return {Array.<Object>}
+        *
+        */
+        this.getUniqueValues = function(attribute, sel){
+            if (typeof (sel) === "undefined")
+                sel = {};            
+            var atoms = getAtomsFromSel(sel);
+            var values = [];
```
dkoes: Don't use lists for membership testing.  Each test is linear.  Use a dictionary.

helblazer811: so make values a dictionary with the structure {'A':true,'B':true} and use hasOwnProperty instead of includes?

PR #321, opened 2017-07-04
Integrated trajectory files parsing through MDsrv

Manohar-94:


3Dmol/3dmol.js
```diff
@@ -152,11 +152,12 @@ $3Dmol.viewers = {};
...
  * @param {string} uri - location of data
  * @param {Function} callback - Function to call with arraybuffer as argument.  
-
+ * @param {string} req - request type
  */ 
-$3Dmol.getbin = function(uri, callback) {
+$3Dmol.getbin = function(uri, callback, req) {
+    req = (req != undefined)?req:'POST';
     $.ajax({url:uri, 
-        type: "GET",
```
dkoes: You can probably just change this to always be POST

3Dmol/glmodel.js
```diff
@@ -2444,6 +2444,34 @@ $3Dmol.GLModel = (function() {
...
+
+    /**
+    * Set coordinates for the atoms parsed from various topology files. 
+    * @function $3Dmol.GLModel#setCoordinatesFromUrl
+    * @param {string} url - contains the url where mdsrv has been hosted
+    * @param {string} path - contains the path of the file (<root>/filename)
+    * @param {object} viewer - contains the glviewer object
+    */
+
+        this.setCoordinatesFromUrl = function(url, path, viewer) {
```
dkoes: setCoordinates should not animate the model.  It definitely shouldn't set the style. The user should be able to call animate using the standard api.  It should setup data structures appropriately so that the coordinate data can be fetched _as needed_ when the frame position is changed in `setFrame`.  There is no need to pass the viewer.

3Dmol/glmodel.js
```diff
@@ -2468,8 +2496,10 @@ $3Dmol.GLModel = (function() {
                     console.log(err);
                 }
             }
-            if (format == "mdcrd" || format == "inpcrd" || format == "pdb" || format == "netcdf") {
-                frames = [];
+            var supportedFormats = ["mdcrd","inpcrd","pdb","netcdf","unknown"];
```
dkoes: As a general rule of thumb, don't use lists for membership testing as it is O(n).  If you are going to query a data structure, make them keys in an object.

tests/auto/tests/test93.js
```diff
@@ -0,0 +1,7 @@
+
+$.get("volData/model1.prmtop",  function (data){
+    var m = viewer.addModel(data, "prmtop");
```
dkoes: you should have to call animate to see it move and should be able to set the style (cartoon would be good, since otherwise you only see water).  Also, need to set the test up so it isn't using a localhost address.  If there is a public mdserv demonstration server, use that, otherwise we can set one up on 3dmol.csb.pitt.edu

3Dmol/3dmol.js
Manohar-94: Done

3Dmol/glmodel.js
```diff
@@ -2462,11 +2462,8 @@ $3Dmol.GLModel = (function() {
                     for (var i = 0; i < numFrames; i++) {
                         $3Dmol.getbin("http://"+url+"/traj/frame/"+i+"/"+path, function (buffer) {
                             self.setCoordinates(buffer, "unknown");
-                            viewer.setStyle({},{sphere:{}});
-                            viewer.zoomTo();
-                            viewer.animate({loop:"forward",reps:1});
-                            viewer.render();
-                        },'POST');
+                            self.setFrame(i)
```
dkoes: setFrame should not be called from within setCoordinatesFromURL

The user called setCoordinatesFromURL _or_ setCoordinates and then they can call setFrame or animate or any other function that changes the frame and it should work as expected.  If the coordinates were initialized from a URL, the code will dynamically fetch the coordinates on demand.  Initially this can be implemented as a synchronous request, but should eventually be implemented with generators (https://eladnava.com/write-synchronous-node-js-code-with-es6-generators/).

3Dmol/3dmol.js
```diff
@@ -157,6 +157,7 @@ $3Dmol.getbin = function(uri, callback) {
     $.ajax({url:uri, 
         type: "POST",
         dataType: "binary",
+        async: "false",
```
dkoes: This can't be left in.  Add an optional argument (default to async) to enable synchronous communication.

3Dmol/glmodel.js
```diff
@@ -2447,34 +2457,40 @@ $3Dmol.GLModel = (function() {
...
     * Set coordinates for the atoms parsed from various topology files. 
-    * @function $3Dmol.GLModel#setCoordinatesFromUrl
+    * @function $3Dmol.GLModel#setupFrames
     * @param {string} url - contains the url where mdsrv has been hosted
     * @param {string} path - contains the path of the file (<root>/filename)
-    * @param {object} viewer - contains the glviewer object
     */
 
-        this.setCoordinatesFromUrl = function(url, path) {
+        this.setupFrames = function (url, path) {
```
dkoes: I prefer the old name.  If data is being hosted remotely, I think we should assume it likely won't fit in memory and we should pre-create frames.  Instead they should be created dynamically, on-the-fly.  This should do the minimum amount of work to provide the information to make setFrame work.

3Dmol/glmodel.js
```diff
@@ -2447,34 +2457,40 @@ $3Dmol.GLModel = (function() {
...
             var atomCount = atoms.length;
             var self = this;
             frames = [];
             $.get("http://"+url+"/traj/numframes/"+path, function (numFrames) {
```
dkoes: This is asynchronous, so if the next call (after setCoordinatesURL) is to setFrame, it won't work.

3Dmol/glmodel.js
```diff
@@ -1446,14 +1446,24 @@ $3Dmol.GLModel = (function() {
...
             }
-            if (framenum >= 0 && framenum < frames.length) {
-                atoms = frames[framenum];
-            }
-            else {
-                atoms = frames[frames.length-1];
+            if (framenum < 0 || framenum > frames.length) {
+                framenum = frames.length - 1;
+            }
+            atoms = frames[framenum];
```
dkoes: Each frame should be dynamically setup when the frame is requested.  Or even better, the coordinates should be read into the existing atoms.  This means the check for the url should come first, although you still need to check the number of frames.

tests/auto/tests/test93.js
```diff
@@ -3,7 +3,8 @@ $.get("volData/model1.prmtop",  function (data){
     var m = viewer.addModel(data, "prmtop");
     var url = "127.0.0.1:800/mdsrv";
     var pathToFile = "root/3Dmol.js/tests/auto/volData/model1_md2.nc";
-    m.setCoordinatesFromUrl(url,pathToFile);
+    jQuery.ajaxSetup({async:false});
```
dkoes: No user should have to do this.

3Dmol/3dmol.js
Manohar-94: done

3Dmol/glmodel.js
Manohar-94: done

3Dmol/glmodel.js
Manohar-94: done

3Dmol/glviewer.js
```diff
@@ -2265,9 +2265,12 @@ $3Dmol.GLViewer = (function() {
             var mostFrames = 0;
             var modelNum = 0;
             for (var i = 0; i < models.length; i++) {
-                if (models[i].getFrames().length > mostFrames) {
```
dkoes: Let's get rid of a `getFrames` as a public api and replace it with `getNumFrames`.  We shouldn't be exposing these implementation details.  Also, change the viewer to be `getNumFrames` as well.

3Dmol/glmodel.js
```diff
@@ -2444,12 +2455,34 @@ $3Dmol.GLModel = (function() {
...
+    * @param {string} path - contains the path of the file (<root>/filename)
+    */
+
+        this.setCoordinatesFromURL = function (url, path) {
+            var atomCount = atoms.length;
+            var self = this;
+            frames = [];
+            $.get("http://"+url+"/traj/numframes/"+path, function (numFrames) {
+                if (!isNaN(parseInt(numFrames))) {
+                    frames.push(atoms);
```
dkoes: This being asynchronous isn't causing a problem?

Manohar-94: It is. I can only render after toggling breakpoints in javascript debugger.

dkoes: It needs to be fixed before it can be merged.

3Dmol/glmodel.js
```diff
@@ -2455,25 +2455,31 @@ $3Dmol.GLModel = (function() {
...
-            if (!isNaN(parseInt(numFrames))) {
-                frames.push(atoms);
-                frames.numFrames = numFrames;
-                frames.url = url;
-                frames.path = path; 
-            }
-
-            yield this.setFrame(0);
-
+            var self = this;
```
dkoes: If we're using promises, then the function should return a promise.  It should not take a callback argument.  The user does something like:
```
setCoordinatesFromURL(url).then(callback);
```

PR #327, opened 2017-07-28
A few bug fixes

helblazer811:
- No more unnecessary scrolling
- Added background color for some buttons
- Changing the model type does not cause any unexpected behavior
- Unnecessary canvases are no longer generated
- Different encoding should not be a problem now
- The dead url works now
- fixed colorschemes


helblazer811:
The issue with the url that wasn't working was that outline is not defined as a valid spec for cartoon. I can add it, but it was just not in the documentation.

PR #328, opened 2017-07-29
Removed runGenerator

Manohar-94:
Changed setInterval to setTimeout in animate


3Dmol/glviewer.js
```diff
@@ -2396,7 +2417,7 @@ $3Dmol.GLViewer = (function() {
                 var newModel = new $3Dmol.GLModel(models.length, defaultcolors);
                 newModel.setAtomDefaults(modelatoms[i]);
                 newModel.addFrame(modelatoms[i]);
-                newModel.setFrame(0);
+                $3Dmol.runGenerator(newModel.setFrame(0));
```
dkoes: shouldn't have runGenerator anymore, right?

3Dmol/glviewer.js
```diff
@@ -2295,34 +2299,51 @@ $3Dmol.GLViewer = (function() {
...
             var displayMax = mostFrames * reps;
+            var time = new Date();
+            var resolve = function() {
+                that.render();
+                if (++displayCount == displayMax || !that.isAnimated()) {
+                    clearTimeout(intervalID);
+                    decAnim(); 
+                }
+                else {
+                    var newInterval = interval - (new Date().getSeconds() - time.getSeconds())/1000000.0;
```
Manohar-94: I was trying to mimic setTimeout to be the same setInterval. In setInterval, display would get called after certain interval regardless of how long it was taking to render whereas in recursive setTimeout, the interval would start getting counted only after the frame had been rendered. I did not want it to wait for some interval if the rendering time had already exceeded the interval.

3Dmol/glviewer.js
Manohar-94: Yes. My bad. I overlooked that.

3Dmol/glviewer.js
dkoes: Buy why divide by 1,000,000?  interval is in milliseconds.

Manohar-94: Oh yeah, I was supposed to use getTime and divide by 1000 to get seconds. Instead of multiplying and cancelling out the 1000, I divided again. Fixed the issue.

PR #331, opened 2017-08-01
implemented promises in getbin and download

Manohar-94:
modified test39.js for testing purposes. All test cases work.


3Dmol/3dmol.js
```diff
@@ -153,19 +153,25 @@ $3Dmol.viewers = {};
...
+    return new Promise(function(resolve, reject) {
+        callback = (callback == undefined)?function(ret){}:callback;
+        request = (request == undefined)?"GET":request;
+        $.ajax({url:uri, 
+            type: request,
+            dataType: "binary",
+            responseType: "arraybuffer",
+            processData: false}).done(
+                function(ret, txt, response) {
+                    callback(ret);
```
dkoes: You can do it this way, but would get more robust error handling if the callback were added to the promise.  Also, creating a dummy functions is a bit inefficient compared to 
```
if(callback) return promise.then(callback);
else return promise;
```

PR #332, opened 2017-08-04
Proof of concept for multi viewer

helblazer811:
I have a split viewer that works. It uses the same mouse movements for all similar to linkedViewer. Some things don't work correctly like backgroundColor.

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -633,6 +633,7 @@ $3Dmol.GLViewer = (function() {
             renderer.setViewport();
             renderer.setClearColorHex(c.getHex(), a);
             show();
+            
             return this;
         };
         
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -92,6 +92,7 @@ $3Dmol.GLViewer = (function() {
         var cols = config.cols;
         var rows = config.rows;
         var viewers = config.viewers;
+        var control_all = config.control_all;
 
         var ASPECT =renderer.getAspect(WIDTH,HEIGHT);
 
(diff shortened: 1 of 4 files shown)


helblazer811:
Do not merge this yet i should be able to add a few new features soon

helblazer811:
This is actually still bugged so ignore this

3Dmol/glviewer.js
```diff
@@ -1202,6 +1206,32 @@ $3Dmol.GLViewer = (function() {
...
+            if(Array.isArray(axis)){
+               axis = {x:axis[0],y:axis[1],z:axis[2]} 
+            }
+
+            var viewer = this;
+
+            
+            spinInterval = setInterval(
+                function(){
+                    console.log("rotating")
```
dkoes: need to remove debug code

3Dmol/glviewer.js
```diff
@@ -1202,6 +1206,32 @@ $3Dmol.GLViewer = (function() {
             ret.normalize();
             return ret;
         };
+        var spinInterval;
+        this.spin = function(spin,axis,interval){
```
dkoes: I was thinking of overloading the first argument to serve the purpose of both spin and axis.
```
spin() 
spin(true)
spin('y')
```
would all do the same thing
`spin(false)` deactivates

helblazer811: I was also thinking that the axis could define speed as well [2,0,0] would spin twice the default on the x axis as [1,0,0]. With just supplying the interval you can only specify one speed for all of the axis and with a high interval there is visible skipping. This way it is continuous and you can specify different speeds for all 3 axis.

PR #333, opened 2017-08-09
implemented promises in addSurface

Manohar-94:


3Dmol/glviewer.js
```diff
@@ -3360,46 +3358,55 @@ $3Dmol.GLViewer = (function() {
...
                     }
+                    var promises = []
+                    for (i = 0; i < extents.length; i++) {
+                        promises.push(startWorker(i));
+                    }
+
+                    return Promise.all(promises)
+                    .then(function(event_data) {
+                        rfunction(event_data);
+                        surfobj.done = true;
```
dkoes: Let's have the returned promise provide the surface id so callers can call `then` on the returned promise and easily access the created surface.

3Dmol/glviewer.js
```diff
@@ -3314,34 +3312,34 @@ $3Dmol.GLViewer = (function() {
...
-                            var mesh = generateSurfaceMesh(atomlist, VandF, mat);
-                            $3Dmol.mergeGeos(surfobj.geo, mesh);
-                        }
-                        _viewer.render();
-
-                        setTimeout(callSyncHelper, 1, i + 1);
+                            resolve();
+                        })
+                        .then(function() {
+                            _viewer.render();
```
dkoes: I don't think all the extents will be guaranteed to be completed when the promise resolves.  Am I wrong? (Also should resolve with the surface id as below).

3Dmol/glviewer.js
Manohar-94: Since, I am using Promise.all, I am not sure how I can change here to return only the surfid, instead of an array of surfid. So, I changed the promise at the end to resolve using surfid.

3Dmol/glviewer.js
Manohar-94: Yes. You are right. I changed it.

3Dmol/glviewer.js
dkoes: You need to return the surfid from the final promise: https://stackoverflow.com/questions/27715275/whats-the-difference-between-returning-value-or-promise-resolve-from-then

Manohar-94: Done.

Issue #285, opened 2017-03-14
Tiled View


pablowatson:
Apologies in advance as this might be a few issues in one.

I was wondering if there are any plans to implement a tiled view (I can't be the first person to have asked for this, though I didn't see anything that looked like this in the issues).

I am in the process of implementing a tiled viewer using 3Dmol.js and was able to link the viewers' rotation and orientation together using the linkViewer function, which I was really happy to discover. It might be work linking the slab modification i.e. ctrl + right button + drag to be linked as well.

This works really nicely and I can add or remove viewers to display whatever the user wants. The problems start when there are say 4 viewers and each contains a large-ish molecule with a surface. If I collapse the tiled view down to one viewer (or vice versa, create a tiled view from from one viewer that contains 4 large-ish molecules with surfaces) I can pretty reliably crash the  browser. I assume this is to do with regenerating the 4 surfaces. I did notice some references to web workers in the source so I assume the surface generation is multi threaded. Is there some way of doing this sequentially, perhaps a callback that notifies when the surface generation is complete.

Also, am I right in assuming you cannot hide/show surfaces - you can only add or remove them.

Sorry for the large list of questions.
Cheers,
Paul

helblazer811:
You can hide/show surfaces with setSurfaceMaterialStyle, which takes a surface style spec, by specifying the opacity to be zero.

dkoes:
You can set `$3Dmol.syncSurface` to true to render surfaces sequentially rather than in parallel (should probably rename this to be more descriptive..).

I'm not entirely sure what your vision is for the tiled viewer that is different than having multiple views linked together?

pablowatson:
Thanks for the suggestions. I will try out the syncSurface setting and setting the opacity of surface to 0 to hide them.

I was hoping for a single canvas/glviewer that would do the tiling. You load the objects in to the viewer, define the tile layout  (e.g. 2 x 2) and say which tile you want it each object to appear in. Each 'tile' is really represents a translation matrix that would place the objects assigned to that 'tile'. Would something like that actually be feasible?

dkoes:
Each tile is the exact same object, or are they different objects?

pablowatson:
The tiles doesn't really exist. 

Maybe you have a map of tile locations to translation matrices e.g. assuming I have a 2 x 2 layout.

tile[0,0] -> matrix that moves the molecules/surfaces etc. to this tile to the top-left
[...]

PR #334, opened 2017-08-14
Bug Fixes 

helblazer811:
Background color for split viewer works now. This should close issue #285 and builds upon #332. Overrode the first parameter for spin. You can now supply only an array, axis or dictionary. true defaults to 'y'. Interval is no longer there; it now defaults to 25. Speed is defined by the axis such that [2,0,0] moves twice the speed on the x axis as [1,0,0].


3Dmol/glviewer.js
```diff
@@ -1207,29 +1206,29 @@ $3Dmol.GLViewer = (function() {
             return ret;
         };
         var spinInterval;
-        this.spin = function(spin,axis,interval){
```
dkoes: Needs documentation.

PR #356, opened 2018-01-25
Change to relevant example in documentation for createViewer

danielholmes:


dkoes:
This was actually more difficult than it appears.  The auto test system generates tests from the embedded examples, but it creates the viewer, so we weren't able to actually test functions that create a new viewer.  I've added code to deal with this case as well as additional test cases/documentation for other viewer creation routines.

PR #375, opened 2018-06-22
Remove unused this assignment

cjh1:
The value doesn't seem to be used and causes problems when running in a ES module as 'this' will be undefined.


dkoes:
I've moved the num to the $3Dmol namespace, documented it, and am actually using it instead of numbers now.
c706390eafb93a57de9331ac866f1d26b7405e65

PR #399, opened 2019-03-24
Add WebGL lost context handling

sjaustirni:
Makes 3Dmol canvases which lost context reload on user mouse click.

Only a demo now, comments (on structure and else) welcome.

### DIFF
--- a/3Dmol/autoload.js
+++ b/3Dmol/autoload.js
@@ -130,9 +130,11 @@ function load(self, viewer = null, nviewers = 0) {
         window.location = "http://get.webgl.org";
     }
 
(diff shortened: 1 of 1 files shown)


sjaustirni:
You can test it on the `test/multi.html` by clicking on the canvases with lost branches.

I haven't realized autoloading was not the only way of using 3Dmol. I assume the general way would be to incorporate it to `$3Dmol.GLViewer`?  I assume this would also take care re-loading the data.
We absolutely need to create a new canvas for it though, if we want to have more than 16 canvases on one page at a time.

Sorry for the late response, I didn't find time to program for the last two days :/

Issue #166, opened 2016-04-11
per-atom opacity


dkoes:
Currently opacity is per-model, which means we don't have to copy a full array of usually identical opacities to the shader.  Should refactor the shaders (with #defines) so there are single-opacity and per-vertex/object (for imposters) opacities and use the appropriate one when opacities are set.

Should probably do something similar for color..

bassemhermina:
just making sure, so for example the effect you're looking for is being able to change this uniform "opacity" with a varying one using a directive #impostervaryingopacity when the input for the opacities for the imposter object is an array of values; not a single value.
![image](https://user-images.githubusercontent.com/23088046/55202209-286fb380-51cf-11e9-8915-a599d474370b.png)

bassemhermina:
I mean is the (varying opacity) input coming in an array from the data like in this form
![image](https://user-images.githubusercontent.com/23088046/55202568-9ff21280-51d0-11e9-9735-eafd15f9995a.png)

dkoes:
Basically opacity should be treated like color is being treated.  It can already be set differently on individual atoms as part of the style.

PR #400, opened 2019-03-31
Added 'per-atom opacity' functionality

bassemhermina:
solving issue #166 
there was a problem with the blank render path, I can't find a proper way without hacking through the legibility of the code to get it to render with opacity = 0 (when the varying opacity is enabled the opacity is not a single value but it's a buffer, and the shader is already built this time so I don't get to have the option of using the single value uniform opacity in the shader) so the only solution that I had in mind was to create another buffer with blank opacities (0,0,0,0,..) through the whole array but that would be a waste.
the other solution/trick that I used is that in the shader I multiply then divide the varying opacity value with the uniform one (ex: opacity_value = varying_opacity * uniform_opacity / uniform_opacity) , that has 2 outcomes:
1- when uniform_opacity is a non zero value, it's canceled and becomes * 1
2- when it's zero (blank render path) , it cancels the varying_opacity and the outcome is 0


bassemhermina:
I created the test file, which has let me see some error in the visual output, the objects which have multiple opacities are rendered in the "transparent" render path which makes them not visible vehind other transparent objects.
So a work arround would be to have multiple opacity buffers, each for each render path, and that would solve the blank render problem too.

dkoes:
Overlapping transparency is a separate issue.

dkoes:
Can you explain what you mean by "blank rendering path".  I still don't get it.

bassemhermina:
Ohh, I'm finishing the final details of the multiple opacity buffers work arround. Will push it in an hour or so.

bassemhermina:
Yes, the  'renderObjects' function have 3 variants , regarding the materialType parameter. 
And when it gets the materialType as 'blank' , it sets the opacity value to 0
But the single valued opacity is not used when the object have multiple opacities/styles and the used value is the opacity buffer

bassemhermina:
Aparently the technique I've followed have failed, because opaque objects have to be rendered before the transparent ones, but having an object which is both opaque and transparent causeed it to be wrongly rendered, so apparantly I'll have to divide the object into smaller entities like what has been done to 'cartoon' objects!

bassemhermina:
> Could you add one more test that uses a small molecule rendered with spheres where a few individual atoms are made transparent (e.g., take caffeine and make only nitrogens transparent). That will make the feature more obvious.
- Sure!

> But you don't see the opaque atoms through the transparent ones.
- that’s what I meant by my comment when I mentioned that the a multiopacity molecule should be divided into smaller ones with a contant opacity, to be rendered individually, may be that can be put into another issue.

dkoes:
What I would do is preprocess the geometries so they are labeled as fully opaque, fully transparent, or mixed.  For the mixed geometries, the shader is run twice - the first time non-opaque triangles are ignored and the second opaque triangles are ignored.

More efficient would be to segment the geometries into opaque and transparent parts, but that is more involved.

bassemhermina:
Yes exactly, I thought about that second solution, as the first would face some problems because of the arrangement of render calls with other molecule objects; but there isn't enough time to finish it before GSoC deadline besides I have midterm exams ongoing right now, so what do you think?

dkoes:
Yes, prioritize the proposal.

bassemhermina:
Ok, great! I'll get back to the issue in a week or 10 days, see you in GSoC :V haha

dkoes:
Why closed?

bassemhermina:
The approach was totally wrong so I reset the fork and deleted all the edits in my commits to start fresh.

PR #401, opened 2019-04-01
Pretty mdpocket.html

nd-02110114:
This is the first PR in this repo.
When I read html files in example dir, I found the iregular indent. 
So, this PR fixed it.  Please confirm 🙇


dkoes:
There isn't anything that needs to be fixed in the code.

Issue #402, opened 2019-04-01
Allow support for saving molecule's 2D image from viewer.html


K7S3:
Currently, the viewer(http://3dmol.csb.pitt.edu/viewer.html) does not have an option to save the molecule as a png/jpg file.
Why is this useful? People who would want to use these images on their slides/ documents would find it very helpful to have it as a PNG file.

PR #408, opened 2019-04-05
Added functionality 'save as PNG' in viewer.html

K7S3:
Solves issue #402 
Remove unnecessary *Render* button.
Added *Save as PNG* option. 
Screenshots:
3dmol(2)
Screenshot from 2019-04-05 22-28-49

### DIFF
--- a/npm-debug.log
+++ b/npm-debug.log
@@ -1,0 +1,11035 @@
+78039 silly preinstall rc@1.2.8 /home/keshavan/3Dmol.js/node_modules/.staging/rc-0185acd3
+78040 info lifecycle rc@1.2.8~preinstall: rc@1.2.8
+78041 silly preinstall path-exists@2.1.0 /home/keshavan/3Dmol.js/node_modules/.staging/path-exists-eb1827ff
+78042 info lifecycle path-exists@2.1.0~preinstall: path-exists@2.1.0
+78043 silly preinstall find-up@1.1.2 /home/keshavan/3Dmol.js/node_modules/.staging/find-up-c0f7f6d2
+78044 info lifecycle find-up@1.1.2~preinstall: find-up@1.1.2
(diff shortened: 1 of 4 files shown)


K7S3:
@dkoes The first part that is 'save as PNG' can be merged, I don't think there are any problems with that.
The second part "documentation to math routines": To check this view this file instead of going through every commit: https://github.com/3dmol/3Dmol.js/blob/40643f3059538ecb74dfe88d04523a13ca68f868/math-routines.md

K7S3:
@dkoes do we need to have separate files of documentation for each class or just having all of them in one is better?

K7S3:
@dkoes can you send an example(existing documentation)

dkoes:
Look at any public function in the existing code.  For example:
```
        
        /**
         * Set lower and upper limit stops for zoom.
         * 
         * @function $3Dmol.GLViewer#setZoomLimits
         * @param {lower} - limit on zoom in (positive number).  Default 0.
         * @param {upper} - limit on zoom out (positive number).  Default infinite.
         * @example
          $.get("data/set1_122_complex.mol2", function(moldata) {
                var m = viewer.addModel(moldata);
                viewer.setStyle({stick:{colorscheme:"Jmol"}});
                viewer.setZoomLimits(100,200);
                viewer.zoomTo();
                viewer.zoom(10); //will not zoom all the way
                viewer.render();
            });
     */
        this.setZoomLimits = function(lower, upper) {
            if(typeof(lower) !== 'undefined') config.lowerZoomLimit = lower;
            if(upper) config.upperZoomLimit = upper;
            rotationGroup.position.z = adjustZoomToLimits(rotationGroup.position.z);
            show();
        };
```

K7S3:
@dkoes I will change it accordingly

PR #410, opened 2019-04-06
Remove unused file

nd-02110114:
`viewer directory` isn't maintained, and the files in this dir are duplicated  in the root.
So, I deleted these files in order to avoid confusing.

In additon to this, I deleted `sample.html` and `doc.js` which are not maintained.


K7S3:
The viewer file in the viewer directory is a new prototype to the existing viewer. However the work is unfinished. If by any chance one tries to implement a full featured viewer (issue #3) the existing work done in this regard might be useful. Once issue #3 is solved, we can remove these files.

PR #412, opened 2019-04-16
V3000 support

pablowatson:
Hi there,

We recently had the need to support V3000 version mol files in 3Dmol.js. So we added the capability to parse this format to parsers.js.

The majority of the work was done by Matt Stahl (github username mattstahl00). I just merged it into the 3Dmol.js project and tidied up the code a bit.

Let me know what you guys think?


pablowatson:
Hi there, I added some tests that load the same V2000 and V3000 molecule and check that it gets rendered on using the image comparison (that is quite a neat way of testing something like 3Dmol.js). Let me know what you think.

PR #414, opened 2019-04-23
fix promise callback breaking py3dmol update method

gravitypersists:
The [js Promise api](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise), which `$3Dmolpromise` is, doesn't have a `done` method (unlike jquery's `$.Deferred`).

Without this fix, calling `view.update()` from a notebook will create an error that `done is not a function`.

This fixes that.


gravitypersists:
Could be [this change here](https://github.com/gravitypersists/3Dmol.js/commit/46cb80653508641a54436197841c97d37be87321#diff-63c4f041d9e62341a652221dc7dbb363L40) which switched the object from a `$.Deferred` to a `Promise`

PR #423, opened 2019-08-22
Volumetric Renderer

bassemhermina:
This is the final commit in the pull request for merging the code done through 3 months of google summer of code (GSoC-2019) for the project "Create Volumetric Renderer module for 3Dmol.js".
The volumetric renderer is added to the final render path (for transparent objects) after all the other objects are rendered, alongside with a custom framebuffer with its depth attachment for depth integration of the volume renderer with the objects rendered.


3Dmol/glviewer.js
```diff
@@ -2575,6 +2576,163 @@ $3Dmol.GLViewer = (function() {
             return s;
         };
         
+        /**
+         * Create volumetric renderer for volumetricData
+         * @function $3Dmol.GLViewer#addVolumetricRenderer
+         * @param {$3Dmol.VolumeData} data - volumetric data
+         * @param {IsoSurfaceSpec} spec - Shape style specification
```
dkoes: I would think this would have its own specification?  For example the transferfn.

bassemhermina: I don't fully understand what you mean by that, can you elaborate on that?

PR #426, opened 2019-08-25
late fixes for merging the volumetric render module

bassemhermina:
*fixed the test that needed files from a different folder
*fixed createGridView errors that caused multiple views to have glitches


bassemhermina:
> Have you tested the case where there are multiple separate webgl contexts on the webpage?

Yes I've, it's working fine.

PR #427, opened 2019-08-26
late fixes for merging the volumetric render module

bassemhermina:
*added FXAA (Fast approximate antialiasing) to screenshader for rendering the custom framebuffer added, as WebGL doesn't support built-in antialiasing to custom FBO


bassemhermina:
So is master fixed after merging?

PR #443, opened 2020-02-20
Use npm to track dependencies on external libraries

NicolasCARPi:
This is a quality of life improvement change for developers.

The bundled jQuery library had a security vulnerability:
https://snyk.io/test/npm/jquery/3.2.1
Also, removing dependencies from the repo is always a good thing!
mmtf.js doesn't seem to be on npm
and pako_inflate.js looks like it's now called pako but I didn't touch
it yet.

Add the package-lock.json so builds are reproducibles.


dkoes:
Great! Thank you.  I've confirmed that everything builds and all tests pass.

PR #445, opened 2020-02-21
Fix python path in echo.cgi

NicolasCARPi:
It's always better to use /usr/bin/env as the python executable might be
in various places.
Also specify that we want python2 as python3 won't work and nowadays it
is probably the default python interpreter.

### DIFF
--- a/echo.cgi
+++ b/echo.cgi
@@ -1,7 +1,6 @@
-#!/usr/bin/python
-#fetchs a file - workaround for cross site scripting prevention
-import cgi,requests,sys
-
+#!/usr/bin/env python2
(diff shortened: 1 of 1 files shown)


dkoes:
I went ahead and updated to python3
68677e0d0607e715f237eeaf59744cda88e6469b

NicolasCARPi:
Good idea! :)

NicolasCARPi:
@dkoes You might want to also update this scripts here too: https://github.com/3dmol/3Dmol.js/tree/master/track
and also set this to python3: https://github.com/3dmol/3Dmol.js/blob/master/.pydevproject

PR #474, opened 2020-09-11
Make a copy of the selection before evaluating them

Luthaf:
This ensures the user can never observe or break the caching done to improve evaluation time

### DIFF
--- a/3Dmol/glmodel.js
+++ b/3Dmol/glmodel.js
@@ -1993,7 +1993,35 @@ $3Dmol.GLModel = (function() {
          */
         this.selectedAtoms = function(sel, from) {
             var ret = [];
-            sel = sel || {};
+
+            // make a deep copy of an object. This does not support arbitrary
+            // javascript objects, but support enough for eveything that is
+            // used in selections: number, string, boolean, functions; as well
+            // as arrays and nested objects with values of the aformentioned
+            // types.
+            const deepCopy = function(object) {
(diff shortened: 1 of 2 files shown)


Luthaf:
This is a follow up to #467

3Dmol/glmodel.js
```diff
@@ -1993,7 +1993,35 @@ $3Dmol.GLModel = (function() {
...
         this.selectedAtoms = function(sel, from) {
             var ret = [];
-            sel = sel || {};
+
+            // make a deep copy of an object. This does not support arbitrary
+            // javascript objects, but support enough for eveything that is
+            // used in selections: number, string, boolean, functions; as well
+            // as arrays and nested objects with values of the aformentioned
+            // types.
+            const deepCopy = function(object) {
```
dkoes, line 2002: when selectedAtoms is recursively called, won't that result in an unnecessary additional copy?  You could create the cache of selectors as part of this function.  If the cache is already created, you know you don't need to do the copy again.

Luthaf: I would indeed result in additional copies of the selection, which I think should be cheap enough to do. 

The issue with creating the caches here is that we would have two places in the code that have to be aware of caches: here and when the actual evaluation takes place. If you think the (small) added code complexity is worth the memory savings, I can implement it 😃

dkoes (approved):

PR #476, opened 2020-10-19
Add the ability to set crystal matrix directly and use it for all unit cell operations

### LINKED ISSUE #469: [question] What's the best way to create GLModel from JavaScript, with unitcell information?
**Description**

I want to create GLModels **without** parsing a file, i.e. I already have the full structure data as JavaScript Object & want to pass these directly to 3Dmol.

I currently do this by creating an empty model & adding atom manually, with 
```js
const viewer = new $3Dmol.GLViewer(...);
const model = viewer.addModel();
model.addAtoms(atoms);
```

My question is how can I then add unitcell/box/CRYST information to the model? I've seen there is a `GLModel.getCrystData`, which seems to look at `modelData.cryst`, containing cell information as `a/b/c/alpha/beta/gamma`. Should I add a public function to `setCrystData`? 

At the same time, using `a/b/c/alpha/beta/gamma` instead of a full cell matrix (three cell vectors/axis) might be problematic since I would have to do a rotation of the entire cell and atoms if the axis are not aligned with `x/y/z`, and potentially loose some information regarding cell orientation (this situation comes up in anisotropic NPT simulation of materials).

My end goal is being able to (a) draw the unit cell box and (b) replicate the unit cell to create supercell, hopefully using `GLViewer.replicateUnitCell`.

**Relevant Context**

For the context, I'm evaluating if we could switch from JSmol to 3Dmol.js in [chemiscope](chemiscope.org/) for molecular rendering. It seems that 3Dmol fits our needs very well, while being easier to use & integrate in a modern JS application. The question above is my main blocking point right now =)

Luthaf:
Fixes #469 

This PR adds the ability to set `cryst.matrix` manually, use it for all unit cell operations and make sure it is set even when the user/parser only specified `a/b/c/alpha/beta/gama`.

### DIFF
--- a/3Dmol/glmodel.js
+++ b/3Dmol/glmodel.js
@@ -1453,7 +1453,7 @@ $3Dmol.GLModel = (function() {
         };
         
         /**
-         * Returns crystallographic information if present.
+         * Set crystallographic information using three angles and three lengths
          *
          * @function $3Dmol.GLModel#setCrystData
          * @param {number} a - length of unit cell side
@@ -2220,7 +2220,6 @@ $3Dmol.GLModel = (function() {
          * 
          * @function $3Dmol.GLModel#removeAtoms
          * @param {type} badatoms - list of atoms
-         * @return {removeAtoms}
          */
         this.removeAtoms = function(badatoms) {
             molObj = null;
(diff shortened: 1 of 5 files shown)


3Dmol/glmodel.js
```diff
@@ -1445,15 +1445,22 @@ $3Dmol.GLModel = (function() {
          */
         this.getCrystData = function() {
             if (modelData.cryst) {
+                // add the matrix if it is missing
+                if (!modelData.cryst.matrix) {
+                    const cryst = modelData.cryst;
+                    modelData.cryst.matrix = $3Dmol.conversionMatrix3(
```
Luthaf, line 1451: Should I rename `conversionMatrix3` to something like `createCrystMatrix` or similar?

dkoes (approved):
Looks good to me.  Thanks!

3Dmol/glmodel.js
dkoes: Nah, that's an even narrower name for a somewhat general function.

PR #488, opened 2020-12-22
Set eye separation value dynamically

camilacanhete:
The function 'setAutoEyeSeparation' doesn't have the option to set the x value. Instead, it sets a fixed value of 5, which may not work for all users/glasses.


dkoes (changes requested):
I think it would be better to add an optional x argument to setAutoEyeSeparation to avoid repetitive code (if it isn't set, default to 5).

Stereo viewing really hasn't been properly evaluated tested.  Any help making this work seamlessly with popular devices (e.g. Google Cardboard) would be appreciated.

dkoes (approved):

Issue #434, opened 2020-01-22
stick imposters have visual artifacts with orthographic projection


dkoes:
it's just ugly

dkoes:
Fixed.

PR #496, opened 2021-01-27
Add a setCameraParameters function to set FOV & CAMERA_Z

Luthaf:
I need to change these parameters as a workaround for #434: using a perspective camera with very small fov and distance to the object feels almost like a perspective camera. 

I tried to make this function generally useful to users of 3Dmol.js, let me know if there are better ways to do something like this!

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -603,12 +603,38 @@ $3Dmol.GLViewer = (function() {
                 viewer.zoom(10); //will not zoom all the way
                 viewer.render();
             });
-     */
+        */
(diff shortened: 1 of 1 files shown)


dkoes (approved):
Looks good to me - thanks!

Luthaf:
Thanks for the quick review and merge! When do you think you'll have time to do a release with these changes? In the mean time I can use git dependencies on the master branch 😃

dkoes:
1.6.2 now released with this change.

Luthaf:
Thanks a lot!

PR #502, opened 2021-02-12
Parsing of unit cell vectors in XYZ files

Mazay0:
Added parsing of the Lattice parameter in extended XYZ files. This parameter contains the unit cell vectors. 
For parsing used regexp from the [ASE library](https://gitlab.com/ase/ase/-/blob/0683eaf282888448427f0951b341bbd1416807a5/ase/io/extxyz.py#L40).


Mazay0:
Here is the file with `Lattice` for testing. Just in case.
[empty-V2CO2.xyz.tar.gz](https://github.com/3dmol/3Dmol.js/files/5973112/empty-V2CO2.xyz.tar.gz)

dkoes:
Thanks!  Merged with a test case.

PR #503, opened 2021-02-16
Addition of speed option for this.spin

BradyAJohnston:
Hi there!

Thanks for the great package. I'm not a javascript developer, so apologies if I'm going about this the wrong way. 

I've added an option for adjusting the speed of the `spin()` function. 

What else do I need to do for this - do I need to add it into other parts of the package? Build some tests for it? Let me know and I'll get on it!


dkoes (approved):

dkoes:
Thanks for the contribution!  The test framework can't test indefinite animations (they have to stop at some point) so I don't think a test case is applicable.

PR #507, opened 2021-03-13
Fix order of lattice vectors coordinates on matrix creation in extxyz…

Mazay0:
Fix order of lattice vectors coordinates on matrix creation in extxyz parser


Mazay0:
Sorry, I don't now how to generate the PNG file for the test. I will do it if you give me some instructions how to do it.

Issue #439, opened 2020-02-10
Volumetric rendering for non-ortho boxes


dkoes:
Current support doesn't work if file format (e.g. ccp4) specifies a transformation matrix.  Shouldn't be too difficult to support since renders take a model to texture transformation matrix already.

jaxmatrix:
I am working on this issues, please tell me where I can find a `dx data file` and `ccp4 data file` with non-orthogonal boxes definition

dkoes:
One of the reasons this issue remains open is I have yet to see an example in the wild.  You can search the PDB for entries that have non-right angles (https://www.rcsb.org/search?request=%7B%22query%22%3A%7B%22type%22%3A%22group%22%2C%22logical_operator%22%3A%22and%22%2C%22nodes%22%3A%5B%7B%22type%22%3A%22group%22%2C%22logical_operator%22%3A%22and%22%2C%22nodes%22%3A%5B%7B%22type%22%3A%22group%22%2C%22logical_operator%22%3A%22and%22%2C%22nodes%22%3A%5B%7B%22type%22%3A%22terminal%22%2C%22service%22%3A%22text%22%2C%22parameters%22%3A%7B%22operator%22%3A%22less%22%2C%22negation%22%3Afalse%2C%22value%22%3A90%2C%22attribute%22%3A%22cell.angle_alpha%22%7D%2C%22node_id%22%3A0%7D%5D%7D%5D%2C%22label%22%3A%22text%22%7D%5D%2C%22label%22%3A%22query-builder%22%7D%2C%22return_type%22%3A%22entry%22%2C%22request_options%22%3A%7B%22pager%22%3A%7B%22start%22%3A0%2C%22rows%22%3A100%7D%2C%22scoring_strategy%22%3A%22combined%22%2C%22sort%22%3A%5B%7B%22sort_by%22%3A%22score%22%2C%22direction%22%3A%22desc%22%7D%5D%7D%2C%22request_info%22%3A%7B%22src%22%3A%22ui%22%2C%22query_id%22%3A%229d8dc3072dabf611c8059c6aefc04ee2%22%7D%7D) and you will find some examples (https://www.rcsb.org/structure/6bna).

However, this reveals a more important issue - the PDB uses DSN6 files (http://www.uoxray.uoregon.edu/tnt/manual/node104.html) which don't yet support.

jaxmatrix:
[...]

PR #511, opened 2021-04-12
Non ortho fix

jaxmatrix:
I am working on the issue #439 

I have added the matrix required for mapping texture to the non-orthogonal box. For this I have re-calculated the dimension of the box that is going to render the volumetric information and added an additional Matrix4 `textmap` which takes the vectors that are used to define the non-orthogonal box dimension in the volume.js. 

This currently works fine with the .cube format remaining formats are to be tested. I am currently working on to make this work for .dx file formats.

Test file can be found in tests/non-ortho-box.html and that are rendered are added in test/test_structs.

image

image

I have checked the volumetric rendering using ChimeraX visualizer

Following are the images generated from [ChimeraX](https://www.rbvi.ucsf.edu/chimerax/)

image

image

image


dkoes:
Thank you for this contribution.

The box still looks orthonormal? The edges aren't at an angle.
box

In the future, please do not commit whitespace changes with code changes. It makes it much harder to review the code.
The test needs to be put into the test framework

jaxmatrix:
I will keep that in mind to take care of the whitespaces. Actually I did few experiments with the mapping of texture, in which I found that we actually don't need a non-orthogonal box. 

We can have our textmap such that it access the volumetric coordinates in the following fashion
image

Here we can see the oblique cube embedded inside the orthogonal points. 

So when we sample the points inside the orthogonal cube, their coordinate are converted to coordinate of system of vector specified in the volumetric data file. Now so if the point that we are sampling belongs to point volumetric data then it will have coordinates between 0 to 1 and for remaining coordinate it will lie outside this range.

Now the outer box that I have defined has the dimension encapsulating the farthest coordinate of the volumetric data. I am still not complete sure about this though. This could have been a lucky case where this logic worked. I am still working on it to clear my hypothesis.

I will shortly share my observations that made me conclude this.

dkoes:
Yes, you do a coordinate conversion, but when the converted point lands outside the box, no density should be generated so you should end up with a slanted looking box of density.

jaxmatrix:
I think the volume intersecting with the box. I will fix it this week end.

dkoes:
There may have been an issue with the incorporation of the translation matrix.  I've made several improvements to the cube handling code and added your example (with isosurfaces) as a test:
http://3dmol.org/tests/auto/generate_test.cgi?test=testcubemol

The goal should be to have the same test with volumetric rendering. A transfer function ranging from 0.03 to 0.075 should look good.

jaxmatrix:
Okay, Thanks. It will be really helpful. I am also thinking of adding a volumetric cube data for the tests so that we can see how are the vectors transforming with our texture map function. 

Earlier I used the "coordinate value as the color in shader.js" to look up how the vectors are being transformed inside the box.

jaxmatrix:
PS: I wrote a script to create non orthogonal boxes (.cube format) and I am testing the volumetric rendering, secondly for this example I replaced the color with coordinates of the mapped coordinates.

I corrected the problem with points outside the box, but another issue came up, as per the calculation the dimension of the box and the inverse matrix should be same but when I render it the volumetric region is partially clipped. 

Volumetric box
image
image


Volumetric box with manual scaling of textmap matrix (matrix to non-orthogonal coordinates to orthogonal coordinates) 
image
image
image


Original box : 
image

So can you share some reference to the following two lines:
```
texmatrix.makeTranslation(-data.origin.x/xoff,-data.origin.y/yoff,-data.origin.z/zoff);
texmatrix.scale({x:1.0/xoff, y:1.0/yoff, z:1.0/zoff});
```

dkoes:
How is this coming?  Should I wait for it to be finished before making the next point release?

jaxmatrix:
No, I am still working on it, I am new to graphics programming so I might need some time to fix it. I added one more condition in the shader similar to that for `seengood`. There is some scaling issue in the texture mapping matrix that I have introduced in the code.

PR #514, opened 2021-04-27
avoid RangeError: Maximum call stack size exceeded when add click event to a long list of atoms

lidaof:
Hi,
At first his PR would not change anything. I encounter a rare issue that when I am trying to add clickable event to a long long list of atoms, I got this `RangeError: Maximum call stack size exceeded` error see below
image

by changing to a for loop instead of using array.apply seems fixes it.

ref: https://stackoverflow.com/questions/22123769/rangeerror-maximum-call-stack-size-exceeded-why

Thanks in advance for considering this PR.

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -273,9 +273,15 @@ $3Dmol.GLViewer = (function() {
                     let hoverable_atoms = model.selectedAtoms({
                         hoverable : true
                     });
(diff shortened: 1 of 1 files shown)


dkoes (approved):

dkoes:
Thank you for resolving this.  It is surprising to me that there isn't a library function to do this, but as far as I can tell the for loop technique is the most efficient approach.

lidaof:
Thank you @dkoes !

PR #522, opened 2021-06-02
give label a default screenOffset to avoid warning message

lidaof:
Hi @dkoes, recent update of 3dmol.js prints warning message `$3Dmol.Material: 'screenOffset' parameter is undefined.`, 
you can see this warning by run the test file: testlabelscreenoffset.js
by giving a default value to avoid this.  thanks for your time.


dkoes:
Thanks, I've pushed a similar change.

PR #527, opened 2021-07-28
Gsoc UI

jaxmatrix:
I am still working on it I will try to fix as much as possible by tomorrow and will add the documentation properly.


dkoes:
Is this finalized yet?  There is no point in my reviewing something that isn't final.

dkoes:
Some issues I've noticed:
- [ ] icons need tooltips to explain what they are
- [ ] hitting enter on a form should submit changes
- [ ] resize moves selection menu icon
- [ ] menus not lined up with widget that creates them (should act like hierarchical menu)
- [ ] context menu interferes with zooming (wait for mouse up)
- [ ] hitting plus multiple times duplicates text in form

jaxmatrix:
Few more issues : 

- [ ] Missing All selector
- [ ] Selection update does not work properly 
- [ ] Bug in removal of selection 
- [ ] Missing check and cross for adding the label 
- [ ] Surface Style can be improved 
- [ ] Removal of surface 'for' and surface 'of' options

jaxmatrix:
Fixed Following Issues


-  hitting enter on a form should submit changes
-  resize moves selection menu icon
-  menus not lined up with widget that creates them (should act like hierarchical menu)
-  context menu interferes with zooming (wait for mouse up)
-  hitting plus multiple times duplicates text in form
-  Missing All selector
-  Selection update does not work properly
-  Bug in removal of selection
-  Missing check and cross for adding the label
-  Surface Style can be improved
-  Removal of surface 'for' and surface 'of' options

Remaining 

**Minor** 
- tool tips for all the widgets
- minor fixes in label 
- addition of warning under in correct input 

**Major**
- Using autoload to set the state of the ui

jaxmatrix:
Following issues are also resolved

- tool tips for all the widgets
- minor fixes in label
- addition of warning under incorrect input

dkoes:
- [ ] Some icons are still missing tooltips (e.g. edit and hide buttons in selection)
- [ ] If you use the title attribute on a div to expose tooltips then they won't get truncated by window edges/other elements (and is easier to do).  This would also fix the issue where scrollbars appear due to the tooltip of the surface menu when it is at the right edge of the window.
- [ ] Changing the selection without changing the style does not modify the view
- [ ] opacity should be an option for most styles
- [ ] The style pane will disappear when the mouse leaves and not update any changed styles (the surface panel also doesn't change the viewer if it is dismissed without clicking the checkmark)
- [ ] In the style pane, if there is text in a text entry box and it is selected, typing replacement text does not replace it, as is the expected behavior, but appends to the value (e.g., if the radius is 2, the user selects this and types 4 instead of ending up with a radius of 4 they get 24)
- [ ] The style pane says "Selection Spec"
- [ ] Many colorschemes are missing (e.g. COLORCarbon - there is more than just green)
- [ ] what is the "self" selection in the surface panel? This is not clear
- [ ] Inconsistent icons between selection and surface (surface does not have green/red check/x)
- [ ] Surface opacity should default to 100%
- [ ] If the style edit panel is dismissed (by mousing out) without hitting the green checkmark, additional selections cannot be created.
- [ ] atom labels can't be styled

jaxmatrix:
Noted, I will fix these issues by the end of this week

jaxmatrix:
I have generated the form using validStyleSpecs specified in the GLModel structure so if there is no opacity visible in the form it is due to reflection from the validStyleSpecs definition if needed I can add the opacity option on them but I am not sure whether they will work or not. Since that will later then depend on the definition used to draw them on the screen.

PR #528, opened 2021-08-11
fix vasp parse split bug

mindon:


dkoes:
Thanks!  I fixed this for cube but missed the same problem in vasp.

PR #536, opened 2021-09-01
Use the wheel event to control zoom

Luthaf:
`mousewheel` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event), and firefox now support the standard `wheel` instead of requiring `DOMMouseScroll`

With the previous version, zooming in and out of a viewer on firefox (for example using the small viewer in https://3dmol.csb.pitt.edu/doc/index.html) would also scroll the page since the call to preventDefault was on a different event.


dkoes (approved):
Thanks!

PR #539, opened 2021-11-18
release memory when finished the `addSurface` work.

dumplings:
when use the `addSurface`, glviewer will create 4 workers every time, but ever never `terminate` them. the broswer memory will be group up and more and more, if i call `addSurface` too many times.

I added a function called `releaseMemory` in glviewer.js file#4248, when glviewer.js#4278, surface work is done, call it and release the memory.

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -49,11 +49,11 @@ $3Dmol.GLViewer = (function() {
         var fixed_labels = [];
         var clickables = []; //things you can click on
         var hoverables = []; //things you can hover over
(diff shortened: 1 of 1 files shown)


dkoes:
Thank you for pointing this out and providing a pull request.  Next time please make sure your code builds with gulp (you were missing semicolons).

dumplings:
sorry i really did it. but it not work at my workspace. i found we dont have eslint etc. maybe i can do this next time for our 3Dmol.js. 
 ^_^

dkoes:
`npm install` should install all the dependencies for you.

PR #546, opened 2021-12-20
Add ability to set start frame when animating

andfaulkner:
Adds new "startFrame" argument to viewer.animate function, which specifies what frame to start the animation on.

Note: this is somewhat rough and untested for backwards and back-and-forth. It also may cause problems when trying to start on a frame that's too high, and will likely crash if this is attempted (this might be OK). A codebase maintainer should verify this (and maybe make additional changes) before merging.

### DIFF
--- a/3Dmol/glviewer.js
+++ b/3Dmol/glviewer.js
@@ -3222,7 +3222,7 @@ $3Dmol.GLViewer = (function() {
             var startFrame = typeof options.startFrame === 'number' ? options.startFrame : 0;
             var currFrame = startFrame;
 
-            var numFramesToAnimate = loop === "backward" ? startFrame : loop === "forward";
+            var numFramesToAnimate = loop === "backward" ? startFrame : mostFrames - startFrame;
 
             var inc = 1;
             if (options.step) {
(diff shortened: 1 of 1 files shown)


3Dmol/glviewer.js
```diff
@@ -3220,14 +3218,19 @@ $3Dmol.GLViewer = (function() {
             }
             var mostFrames = this.getNumFrames();
             var that = this;
-            var currFrame = 0;
+
+            var startFrame = typeof options.startFrame === 'number' ? options.startFrame : 0;
+            var currFrame = startFrame;
+
+            var numFramesToAnimate = loop === "backward" ? startFrame : mostFrames - startFrame;
```
dkoes, line 3225: I don't think this is right for "backAndForth" - in that case it should cycle through all the frames ending back at the start frame, so numFramesToAnimate should be mostFrames

dkoes (changes requested):
Looks good other than the behavior with backAndForth looping.  A test case (in tests/auto/tests) would be appreciated (e.g., set startFrame and backwards so the final frame is the first)

PR #564, opened 2022-02-21
Feature eslint

angel4576:
# Description
Installed ESLint and Prettier, and installed eslint-config-prettier to make ESLint and Prettier work together; Set up configuration of ESLint and Prettier. The purpose of this is to have a code formatter and style guide.
# Type of change
New feature (non-breaking change which adds functionality)
# How Has This Been Tested?
`npm run build`
**My Configuration**:
* `npm --version`: 8.3.1
* `node --version`: v16.14.0
* `python3 --version`: 3.10.0
* OS: Windows
* ARCH: x86_64

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,10 @@
+{
+  "env": {
+    "browser": true,
+    "es2021": true
+  },
+  "parserOptions": {
+    "ecmaVersion": "latest"
+  },
+  "extends": ["plugin:prettier/recommended", "standard"]
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,8 @@
     },
     "extends": [
         "airbnb-base",
-        "prettier/prettier:recomended"
+        //"prettier/prettier:recomended"
+        "prettier"
     ],
     "parserOptions": {
         "ecmaVersion": "latest"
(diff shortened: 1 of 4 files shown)


dkoes:
How is this to be run?  When I run `npx eslint 3Dmol/*js` I get thousands of errors, some of which are definitely wrong (e.g. `'$3Dmol' is not defined`).  What is the plan for getting the code into conformity with the style guide and enforcing it?

angel4576:
> How is this to be run? When I run `npx eslint 3Dmol/*js` I get thousands of errors, some of which are definitely wrong (e.g. `'$3Dmol' is not defined`). What is the plan for getting the code into conformity with the style guide and enforcing it?

Hello professor, the current configuration is that the eslint will format the file every time when the file is saved. This will fix many format errors, but there are still many that will not be automatically fixed. I think we may need to manually fix them based on the rules and this may be the work we are going to do in the current sprint.

dkoes:
Saved how?  Are you using a specific editor that hooks into eslint?  In order to be a general solution, this should be part of the build system, not an editor.

RyanGuild:
eslint --fix ./3Dmol/*js will fix all auto fixable errors

angel4576:
> `eslint --fix ./3Dmol/*js` will fix all auto fixable errors

so that it is editor independent. Sorry that I have confused them

RyanGuild:
closed in favor of #577

PR #566, opened 2022-02-24
Feature Jest Testing

tjakubiec218:
# Description

* This change adds the ability to use Jest unit testing by running npm run jest
* It also adds a Jest GitHub action that runs the Jest tests on push to master and pull requests from any branch

Fixes # 561

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

-[] npm run build

**My Configuration**:
* `npm --version`: 8.5.0
* `node --version`: v17.1.0
* `python3 --version`: 3.9.1
* OS: Windows
* ARCH: x86_64


tests/tests.test.js
```diff
@@ -0,0 +1,48 @@
+/**
+*@jest-environment jsdom
+*/
+
+global.$ = require("jquery");
+global.URL.createObjectURL = function() {};
+let $3Dmol = require("../build/3Dmol-min.js");
+
+console.log($3Dmol.Quaternion);
```
dkoes, line 9: I don't think this print statement is meant to be here.

package.json
```diff
@@ -23,19 +31,22 @@
...
+    "jsdom": "^19.0.0",
     "jshint": "^2.13.1"
   },
   "scripts": {
     "build": "gulp",
+    "prepare": "npm run build",
+    "jest": "npx jest",
     "types": "jsdoc -t node_modules/tsd-jsdoc/dist -r 3Dmol/. -d types",
-    "test": "npm run glcheck",
+    "test": "npm run glcheck && npx jest",
```
dkoes, line 44: Run jest first - it is faster and simpler.

dkoes (changes requested):

js/mmtf.js
```diff
@@ -1,6 +1,8 @@
 (function (global, factory) {
   (factory((global['MMTF'] = global.MMTF || {})));
-}(this, function (exports) { 'use strict';
+
+}(window, function (exports) { 'use strict';
```
dkoes, line 4: I don't think we should hardcode global into being window no matter what.  Instead it might better to set
`global = global || window;`
although what really needs to happen is properly modularize everything.

dkoes (approved):

PR #604, opened 2022-06-22
TS port SpritePlugin, core/Color, ShaderLib, and ShaderUtils

RyanGuild:
typescript port of SpritePlugin, core/Color, ShaderLib, and ShaderUtils


dkoes (approved):
I'm slightly worried about the changes to external sources (e.g. mmtf.js).  In the future we might need to update these to newer versions and merging could become a pain.

PR #615, opened 2022-07-25
moved files around and updated configs

RyanGuild:
I used the git mv command so this should not create conflicts


RyanGuild:
closing as unfinished

PR #623, opened 2022-10-07
Wait for the DOM to be loaded to install jquery transport

Luthaf:
I am trying to use 3Dmol inside jupyterlab, and jquery is not yet available when 3Dmol code is first loaded, which causes a crash. The simple fix is to delay until the DOM is fully loaded before installing the new jquery transport mechanism for binary files. 

Let me know if you think of a better way to do this!

### DIFF
--- a/packages/3Dmol/src/js/main/3dmol.js
+++ b/packages/3Dmol/src/js/main/3dmol.js
@@ -58,55 +58,59 @@ if (!String.prototype.endsWith) {
 *
 */
 
-// use this transport for "binary" data type
-$.ajaxTransport(
-               "+binary",
-               function(options, originalOptions, jqXHR) {
-                   // check for conditions and support for blob / arraybuffer response type
-                   if (window.FormData && ((options.dataType && (options.dataType == 'binary')) || 
-                           (options.data && ((window.ArrayBuffer && options.data instanceof ArrayBuffer) || 
-                                   (window.Blob && options.data instanceof Blob))))) {
-                       return {
-                           // create new XMLHttpRequest
-                           send : function(headers, callback) {
-                               // setup all variables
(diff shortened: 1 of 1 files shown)


dkoes:
This seems to reproducibly cause a test failures.  Unfortunately, I don't have time to look at it in the immediate future.  It seems very strange that jquery wouldn't be defined, since in the built 3Dmol.js file jquery comes before everything else.  Perhaps you could provide a test case that reproduces the original error?

Luthaf:
I'll try to debug this a bit more, I think it is related to how our code (which re-bundles 3DMol) gets loaded by jupyterlab.

Luthaf:
Ok, so what seems to be happening is that the code is build as a CommonJS module in jupyterlab standard build procedure. This means that when loading jquery, we take this branch: https://github.com/jquery/jquery/blob/5d5ea015114092c157311c4948f7cc3d8c8e7f8a/src/wrapper.js#L25

Calling the jquery factory with `noGlobal` set to true. Which in turn ends up around here https://github.com/jquery/jquery/blob/5d5ea015114092c157311c4948f7cc3d8c8e7f8a/src/exports/global.js#L26-L28

and prevents the export of a global `$` symbol. Making the code changed in the PR fail, since it assumes there is such global symbol.

----

This PR just fixed the symptoms (i was re-exporting `$` as a global at a later point, making it visible then), and not the source of the issue, so I'll close it. I also found a workaround on my side, so this is not urgent. 

However, there might be something that can be done from 3Dmol side to help with this, maybe by explicitly checking for `$` and instantiating/grabbing it from the CommonJS module otherwise, or building it inside the 3Dmol bundle with webpack instead of concatenating the pre-built file with gulp. I'll try to come up with a repo which reproduces the error and open an issue.

PR #650, opened 2023-02-21
Small TypeScript changes

Luthaf:
This PR contains multiple small changes to the TS declaration that we need in our usage of 3Dmol as a library. Each commit should be independent, and I'm happy to address each one differently if needed!


dkoes (approved):

dkoes:
Better typing is always appreciated!

Luthaf:
Thanks for the quick merge!

PR #655, opened 2023-03-10
clicksphere style: Sort to the end of intersection list from the raycaster

donbyron02:
We apply the `clicksphere` style to fragment atoms that make up an interactive hot spot.
The atoms are invisible, but we draw a transparent surface around them. The clicksphere style allows the hotspots clouds to be mouse interactive.
We noticed that if there were protein or compound atoms inside the hot spot, it was preferable to have the mouse respond to the visible atoms, instead of the invisible fragment atoms making up the cloud.
This PR sorts the clicksphere atoms to the end of the intersection objects from the raycaster, so that visible atoms will have priority in the hit detection.

### DIFF
--- a/src/GLViewer.ts
+++ b/src/GLViewer.ts
@@ -620,6 +620,32 @@ export class GLViewer {
     };
 
     /**
+     * @description Modify a list of intersection objects beyond the raycaster's distance sort.
+     * Initially, this just sorts clickspheres to the end.
+     * @param {Array<{clickable: any, distance: number}>} intersectsIn
+     *
+     * Use case:
+     * We'd like an interactive cloud (surface) around some atoms, but don't want to see the atoms.
+     * To make the cloud mouse-interactive, the atoms are given a clicksphere style.
+     * Give intersection priority to shapes and visible atoms so that when mousing over a compound
+     * inside a cloud, the interaction is with the compound, not the cloud.
+     */
+    private filterIntersects(intersectsIn: { clickable, distance: number }[]) {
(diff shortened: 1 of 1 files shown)


dkoes:
I'm not sure how I feel about this as it seems like a special case.  I think it is reasonable to expect he most forward object to be returned.  I think a better fix is for the clickHandler to also be passed the list of all intersecting objects (as an additional argument to what is already passed) so if the implementor has some other preferred behavior they can implement it themselves.  What do you think?

dkoes:
Do you need this both hover and clicking, or just clicks?

donbyron02:
Hi Professor Koes, thanks for your message!
Hmm, I wasn't sure at first if I should submit this PR or just leave it in our fork as one of several customizations.  I ended up convincing myself the clicksphere style (a clickable, invisible atom) was itself the special case and it was always undesirable to have an invisible atom blocking a visible atom.  But you're right there could be other applications of clicksphere that would not want this behavior. 

I don't mind at all keeping this change in our fork. That may be a cleaner solution than adding another callback parameter.

To answer your question, we use it for both hover, click, and more recently, contextmenu.

dkoes:
I went ahead and added the whole list as an additional parameter to the callback function.  It should be easy to switch to using that - one less thing to have to worry about diverging from the main branch.

donbyron02:
Thanks, Prof. Koes -  your responsiveness is legendary!

You might also consider making the last argument an object, with the intersectList as a field, in case there are ever any other values that you'd need to return with the mouse events.

PR #656, opened 2023-03-13
Add tolerance for touch clicks

donbyron02:
It can be difficult to keep your finger in exactly the right place when using a touch interface.
This PR adds a `closeEnoughForClick` which allows a delta when comparing the position against the mouseStart.

By default, it requires exact position for mouse and allows tolerance for touch events (events with targetTouches), but the caller can also specify if they want to allow or disallow the tolerance.

This function will be used in a followup PR for contextmenu events.

### DIFF
--- a/src/GLViewer.ts
+++ b/src/GLViewer.ts
@@ -408,16 +408,22 @@ export class GLViewer {
         }
     };
 
-    // Determine if a positioned event is "close enough" to be considered a click event
-    // With a mouse the position should be exact.
-    // But allow some wiggle room when using touch interface.
-    private closeEnoughForClick(event, { allowWiggleRoom=null, tolerance=5}={}) {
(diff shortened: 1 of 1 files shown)


dkoes (approved):

dkoes:
I assume this fixes a problem seen in the wild?  What hardware?

donbyron02:
Ok, good question. I dug into why we did this, and the specific reason was not for clicks, but to support long-press context menu activation (before you had implemented it). We showed the context menu on `touchend` after a long touch, and since a single function handles both mouseup and touchend, we put the fix there.
Now that we can use the built-in long-touch context menu detection, maybe this isn't strictly necessary in the mouseup handler. I will have a follow-up context menu PR that will use this function, so you may prefer to revert this PR and consider the fuller context later. My apologies for getting confused about this.

PR #659, opened 2023-03-23
Added CodeQL Workflow

adithyaakrishna:
**Description:**

This PR adds CodeQL Workflow to the 3DMol.js Repo

**More Info:** https://codeql.github.com/

### DIFF
--- a/.github/workflows/CodeQL-Analysis.yml
+++ b/.github/workflows/CodeQL-Analysis.yml
@@ -1,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '0 */24 * * 1,3,5'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'typescript', 'python' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Build javascript file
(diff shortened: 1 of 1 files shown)


adithyaakrishna:
@dkoes I guess the security alerts are only visible to you with respect to the above comments

CC @NicolasCARPi

.github/workflows/CodeQL-Analysis.yml
```diff
@@ -0,0 +1,41 @@
...
+      with:
+        languages: ${{ matrix.language }}
+
+    - name: Build javascript file
+      run: |
+        npm i
+        npm run build
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
```
NicolasCARPi, line 41: missing a newline at the end! ;)

adithyaakrishna: Oops! My bad, thanks for pointing it out :)

adithyaakrishna: @NicolasCARPi Fixed it with the below commit :)

.github/workflows/CodeQL-Analysis.yml
```diff
@@ -0,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '0 */24 * * 1,3,5'
```
adithyaakrishna, line 10: This job is scheduled to run according to this Cron Schedule,

Screenshot 2023-03-23 at 6 37 44 PM

dkoes (approved):

.github/workflows/CodeQL-Analysis.yml
NicolasCARPi: Why would you run this on a schedule? It makes sense on push or PR but I don't see how a schedule would be useful. If you're thinking of vulns, this is already automatically done by github.

adithyaakrishna: Hi @NicolasCARPi 
Yes, it makes sense to remove the schedule trigger, I have removed it in the new PR #661

Issue #661, opened 2023-03-23
[Feat] - Performance Improvements  For TS Files in src/


adithyaakrishna:
**Is your feature request related to a problem? Please describe.**

The typescript files in the [src/](https://github.com/3dmol/3Dmol.js/tree/master/src) directory needs performance improvements in terms of code and code readability.

- [ ] GLDraw.ts
- [ ] GLModel.ts
- [ ] GLShape.ts
- [ ] GLViewer.ts
- [ ] Gradient.ts
- [ ] Label.ts
- [ ] ProteinSurface4.ts
- [ ] VolumeData.ts
- [ ] VolumetricRender.ts
- [ ] autoload.ts
- [ ] colors.ts
- [ ] glcartoon.ts
- [ ] index.ts
- [ ] partialCharges.ts
- [ ] specs.ts
- [ ] utilities.ts

**Describe the solution you'd like**
I would like to improve the code base of the 3Dmol.js by using the latest JavaScript methods along with increasing the readability of these files

**Note:** I have created this issue to track the file improvements with respect to the above files, and each commit/PR will improve the performance aspects of the code along with increasing code readability

PR #662, opened 2023-03-23
[Feat] - Performance Improvements to `GLDraw.ts`, `GLCartoon.ts`, `GLModel.ts`

adithyaakrishna:
**Description:**

- This PR is Related to #661 - Optimized GLDraw.ts
- This PR is Related to #661 - Optimized GLCartoon.ts
- This PR is Related to #661 - Optimized GLModel.ts


adithyaakrishna:
@dkoes I'm trying to figure out why the CI is failing here, but wasn't able to. Could you please help me with this or suggest what needs to be done?

dkoes:
Is `npm test` not failing in your local checkout?

adithyaakrishna:
@dkoes Figured it out, it's an issue with `.set()` functions which I added for Arrays. They are mostly [Float32Arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Float32Array) and I don't think the `set()` works with them, I will send a fix for that :)

adithyaakrishna:
I think the set() methods are working fine, but its something related to TS and Node/NPM

@dkoes The tests are failing on the master branch for me :(

Around 18 tests are failing not sure why and this is the error which I am getting for all of them,

Screenshot 2023-03-23 at 11 09 13 PM

adithyaakrishna:
@dkoes I found the issue to be my system itself, its based on arm architecture and node-canvas does not support prebuilt binaries for this architecture. I will continue to update the codebase based on perf improvements. I have a secondary PC which used intel x86_x64 architecture in which I can run the tests and check where it is breaking down :)

src/GLModel.ts
```diff
@@ -175,82 +175,80 @@ export class GLModel {
...
+            if (bond !== atom2.index) {
+                const [j2, atom3] = [bond, this.atoms[bond]];
+                const p3 = new Vector3(atom3.x, atom3.y, atom3.z);
+                const dir2 = p3.clone().sub(p1);
+                const v = dir2.clone().cross(dir);
+                const l = v.lengthSq();
                 if (l > bestlen) {
                     bestlen = l;
                     bestv = v;
-                    if (bestlen > 0.1) {
```
dkoes, line 246: You've removed the short circuiting of the loop so now it will run longer.  Why do you think this is an improvement?

src/GLDraw.ts
```diff
@@ -25,56 +30,19 @@ export enum CAP {
...
-        rot[0] = cosA;
-        rot[1] = sinA;
-        rot[2] = 0;
-        rot[3] = -sinA * cosB;
-        rot[4] = cosA * cosB;
-        rot[5] = sinB;
-        rot[6] = sinA * sinB;
-        rot[7] = -cosA * sinB;
-        rot[8] = cosB;
+        const rot = new Float32Array([cosA, sinA, 0, -sinA * cosB, cosA * cosB, sinB, sinA * sinB, -cosA * sinB, cosB]);
```
dkoes, line 43: Setting the values individually instead of allocating a list first is more than twice as fast (go ahead and run it through JSBench.Me)

dkoes (changes requested):
I am never going to approve this PR.  There are too many changes and too many of them would introduce regressions.  I suggest starting over.  Only make a change if you can prove it is actually an improvement and provide the proof in the PR.  There should not be one single "performance improvement" PR, but separate PRs for each issue addressed. (e.g. use of Math.hypot at all relevant places)

adithyaakrishna:
@dkoes I completely understand, I will work on creating seperate PRs for the same. There were a couple of changes I had done, but these were without much understanding of the library. I will close this and create seperate ones as you suggested

PR #665, opened 2023-03-26
Add more folders to .npmignore

NicolasCARPi:
Hello,

Installing `3Dmol.js` with npm results in a 40 Mb folder. Looking into the size distribution, we can see that a lot of space is used by the python stuff:

2023-03-26-140757_1074x654_scrot

And we definitely don't need all these examples and non-production code when installing the package. Especially the notebooks chekpoints!

So here is a PR that will make the npm install lighter.

Here is how it looks before:

pack-master

And after:

pack-npmignore

As a side note, I would suggest splitting the python code into another repository, as I believe it makes more sense to separate JS and Python projects, which have different tooling and environments.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,3 @@
-/tests
 /node_modules
 /tmp
 /coverage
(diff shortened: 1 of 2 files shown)


dkoes (approved):

NicolasCARPi:
Thanks for the rapid merge @dkoes. Now we can test the new github action to publish a new package? :)

PR #677, opened 2023-05-12
TypeScript: LabelsSpec.alignment can also be a Vector2

Luthaf:
This updates `LabelsSpec.alignment` to follow what the code is doing


Luthaf:
The test failure seems unrelated

dkoes (approved):
I suspect the build failure is due to this error: https://github.com/nodejs/node/issues/47563
and will be fixed in the next node update.  Otherwise, looks fine.  Thanks!

dkoes (approved):
I suspect the build failure is due to this error: https://github.com/nodejs/node/issues/47563
and will be fixed in the next node update.  Otherwise, looks fine.  Thanks!

PR #686, opened 2023-06-13
[Feat] - Changed Math.Sqrt Methods

adithyaakrishna:
**Description:**

This PR changes,

  - `Math.sqrt(dx * dx + dy * dy)` to `Math.hypot(dx, dy)`
  - `Math.sqrt(dy * dy + dz * dz)` to `Math.hypot(dy, dz)`


adithyaakrishna:
But why `hypot()` compared to `sqrt()`?

**Benchmark**: https://measurethat.net/Benchmarks/Show/25689/0/hypot-vs-sqrt

adithyaakrishna:
Gonna keep this PR sweet and small, @dkoes If it looks good, can we merge this?

src/GLViewer.ts
```diff
@@ -2784,8 +2784,8 @@ export class GLViewer {
                 var u, v, w;
 
                 u = Math.cos(beta);
-                v = (Math.cos(alpha) - Math.cos(beta) * Math.cos(gamma)) / Math.sin(gamma);
-                w = Math.sqrt(Math.max(0, 1 - u * u - v * v));
```
dkoes, line 2788: why is this changed?  It doesn't calculate the same value and removes the guard against generating a NaN.

dkoes (changes requested):
Only hypot changes please.

src/GLViewer.ts
adithyaakrishna: @dkoes Does u,v accept negative values?

dkoes: Why does that matter?

dkoes (approved):

PR #690, opened 2023-06-14
[Feat] - Updated Sessions.js

adithyaakrishna:
**Description:**

- Used Array.from() instead of .push()

### DIFF
--- a/learning-environment/sessions.js
+++ b/learning-environment/sessions.js
@@ -297,7 +297,6 @@ var initSessions = function() {
     
     //return information about what is currently clicked for the server
     var getClicked = function() {
-        var ret = [];
         const ret = Array.from(clicked_labels, ({ stylespec }) => [stylespec.position.x, stylespec.position.y, stylespec.position.z]);
         return ret;
     };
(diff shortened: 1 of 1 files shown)


adithyaakrishna:
@dkoes Could you pls lmk if these changes looks good?

learning-environment/sessions.js
```diff
@@ -298,10 +298,7 @@ var initSessions = function() {
     //return information about what is currently clicked for the server
     var getClicked = function() {
         var ret = [];
```
dkoes, line 300: need to remove the var ret delaration

dkoes (changes requested):

learning-environment/sessions.js
```diff
@@ -386,22 +383,20 @@ var initSessions = function() {
...
-            for (let i = 0; i < query_result.length; i++) {
-                let pos = query_result[i][0];
-                let p = {x: pos[0], y: pos[1], z: pos[2]};
-                let val = parseInt(query_result[i][1]);                
-                let l = glviewer.addLabel(val, {position : p,backgroundOpacity:0.5+(val/max)*.5});                
-                result_labels.push(l);                
-            }
+        const result_labels = Array.from(query_result, ([pos, val]) => {
+            const p = { x: pos[0], y: pos[1], z: pos[2] };
+            const label = glviewer.addLabel(val, { position: p, backgroundOpacity: 0.5 + (val / max) * 0.5 });
```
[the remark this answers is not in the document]
adithyaakrishna: Oops! My bad, thanks for pointing it out :)

learning-environment/sessions.js
```diff
@@ -386,22 +382,28 @@ var initSessions = function() {
...
-    });
+        
+        // First clear existing labels
+        clearResultLabels();
+        
+        let max = 1;
+        let i = 0;
+      
+        // Calculate max
+        while (i < query_result.length) {
```
dkoes, line 393: why was this changed to a while loop? Using a while loop for simple iteration is poor coding practice. If you want a more modern pattern use forEach.

dkoes (changes requested):

learning-environment/sessions.js
adithyaakrishna: @dkoes I tried benching them now, their performance benefit compared to forEach is a bit on the lower end, I will make the change and send it in sometime :)

dkoes (approved):

PR #695, opened 2023-07-09
Updated Webpack Config Files & Add Citation File

adithyaakrishna:
**Description:**

- This PR updated webpack config to show error details and also has some linting updates.
- I have also added a Citation.cff file such that users can cite this repository in a easier way. 

More Info: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

The citation file adds something like this on the side menu

Screenshot 2023-07-10 at 12 23 42 AM

Screenshot 2023-07-10 at 12 24 16 AM


adithyaakrishna:
I have created the citation file to the best of my knowledge, please do let me know if there are any other changes required to it 😶

.github/workflows/CI.yml
```diff
@@ -17,14 +17,9 @@ jobs:
...
-      - name: Install Dependencies
+      # As our package has a postInstall script, we dont need a explicit job for that
+      - name: Install & Build Dependencies
         run: npm install 
-      
-      - name: Build 3Dmol.js
-        run: |
-          npm run build:dev
-          npm run build:prod
-          npm run generate:tests
```
adithyaakrishna, line 27: Removed the Build job, which ~saves us 30 seconds in CI time

But why this change? Our `package.json` has a `"postInstall"` script here, https://github.com/3dmol/3Dmol.js/blob/master/package.json#L33 which runs `npm run build` upon installation of dependencies. Till now we were running a redundant job, hence removed the same

webpack.common.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-undef*/
```
dkoes, line 1: This is here for a reason - so VS code doesn't report errors where there are none.

PR #712, opened 2023-08-22
chore: Removed Unused Variables

adithyaakrishna:
**Description:**

- Removed unused variables and added `// @ts-ignore` for needed ones

### DIFF
--- a/src/WebGL/Renderer.ts
+++ b/src/WebGL/Renderer.ts
@@ -69,7 +69,7 @@ export class Renderer {
   // GL state cache
   private _oldDoubleSided = -1 as number | boolean;
   private _oldFlipSided = -1 as number | boolean;
-  // @ts-ignore
+  // Remove
   private _oldBlending = -1;
   private _oldDepthTest = -1;
   private _oldDepthWrite = -1;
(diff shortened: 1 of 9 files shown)


src/WebGL/Renderer.ts
```diff
@@ -1797,10 +1799,12 @@ export class Renderer {
       b = 0,
       color,
       intensity,
+      // @ts-ignore
```
dkoes, line 1802: what is the purpose of all the ts-ignore statement?

src/WebGL/Renderer.ts
```diff
@@ -69,6 +69,7 @@ export class Renderer {
   // GL state cache
   private _oldDoubleSided = -1 as number | boolean;
   private _oldFlipSided = -1 as number | boolean;
+  // @ts-ignore
   private _oldBlending = -1;
```
adithyaakrishna, line 73: Regarding, https://github.com/3dmol/3Dmol.js/pull/712#discussion_r1303000686, I wasnt sure if I should go ahead and remove unused variables in those places. Hence added // @ts-ignore. We can discuss it on our call today and finish this one up :)

src/WebGL/Renderer.ts
adithyaakrishna: Just referenced the comment and addressed it here, https://github.com/3dmol/3Dmol.js/pull/712/files#r1303022588

src/WebGL/Renderer.ts
adithyaakrishna: Removed the unused variables and `// @ts-ignore` :)

PR #728, opened 2023-10-14
Include ranges in typing for resi `AtomSelectionSpec`

randyn:
* Adds a literal type for a SelectionRange and adds it into the union type for `resi`
* Adds arrays as a valid type for `resi` as well

There are a couple other fields on `AtomSpec` that are also typed as number, but I'm not sure which ones we intentionally want to allow ranges for. I'm happy to update those too, this just happened to be the only one I'm looking at right onw

### DIFF
--- a/src/specs.ts
+++ b/src/specs.ts
@@ -75,7 +75,6 @@ export interface AtomSpec  {
   contextMenuEnabled?: boolean;
 };
 
-
 /**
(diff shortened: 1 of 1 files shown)


dkoes:
Better typing is always appreciated!  Although maintaining the whitespace style makes for a cleaner diff.

PR #740, opened 2023-12-07
Dashed bonds for bond order < 1 (revised)

dxdc:
Builds on https://github.com/3dmol/3Dmol.js/pull/641

- Previously the dash calculation was only integrated into drawDashedStickImposter. This was inflexible, as there was no native support for dashes in GLDraw.drawCylinder.

- Updated dash calculation algorithm. Seems to result in improved behavior.


dkoes (approved):

dkoes:
I'm not sure what the issue is with coloration - the test cases look like an improvement to me, so I'm going to go ahead and merge.

PR #749, opened 2024-01-18
Remove redundant argument from package.json

### LINKED ISSUE #748: Remove redundant argument from package.json scripts
**Description**
Currently, there are different webpack config files for different modes (dev/prod) with the base config file being `webpack.common.js`. The modes are mentioned in the corresponding config files (`development`/`production`). However, the modes are again explicitly passed when running scripts like `build:dev`,`build:prod`. I think this is not necessary and adds to redundancy. We can simply remove the `--mode` argument from `build:dev` and `build:prod` command.
Reference: [Webpack mode](https://webpack.js.org/configuration/mode/)

prajwalkulkarni:
# :dizzy: Changelog
:star: Removed redundant `--mode` argument from the CLI command in `package.json` 
:star: Fixes #748

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,8 @@
   "types": "build/types/index.d.ts",
   "scripts": {
     "build": "npm run build:dev  && npm run build:prod && npm run generate:tests && npm run doc",
-    "build:dev": "webpack --config webpack.dev.js --mode development",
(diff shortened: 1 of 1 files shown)


prajwalkulkarni:
@dkoes could you kindly review these minor changes?

PR #754, opened 2024-01-21
Add strict equality operator to TS files src/*

prajwalkulkarni:
# :dizzy: Changelog
:star: Updates TS files `src/*` to use strict equality operator.
:star: Only those conditions are updated where it is safe to update `==` to `===` and `!=` to `!==`. (E.g: statements where types of both the operands are known and same, operands that use `typeof` etc.)
:star: There still exist several statements where the operator is not updated as the logic might change (`==` returns true on comparing `null` and `undefined` whereas `===` returns false) or the types of operands are not known (`any`). Updating this requires manual efforts and it'll be easier once the type safety of the library has been improved. 
:star: The operators still need to be updated in these paths: `src/parsers`, `src/WebGL` which will be done on a separate PR (because it'll be difficult to self-review large file changes before submitting PR)

### DIFF
--- a/src/GLDraw.ts
+++ b/src/GLDraw.ts
@@ -224,7 +224,7 @@ export namespace GLDraw {
                             if (Math.abs(vertex.z) < 1e-5)
                                 vertex.z = 0;
 
-                            if (cap == CAP.FLAT) {
+                            if (cap === CAP.FLAT) {
                                 n = new Vector3(0, Math.cos(thetaStart + v * thetaLength), 0);
                                 n.normalize();
                             }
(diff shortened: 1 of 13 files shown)


prajwalkulkarni:
@dkoes Could you kindly review and let me know if there are any other changes to be made?

cc @NicolasCARPi

prajwalkulkarni:
UPDATE: Removed `typeof` wherever a variable/property was compared to `undefined`. Added optional chaining wherever a nested property was compared to `undefined`, e.g.: `a.properties[prop]` in `GLModel.ts` where `properties` is an optional type and could possibly be undefined.

PR #756, opened 2024-01-22
Refactor ts files parser utils

prajwalkulkarni:
# :dizzy: Changelog
:star: Improved typesafety in TS files located in `parser/utils`.
:star: Changed `var` to `let` and `const` as applicable.
:star: Moved constant value identifiers outside of function definition to avoid unnecessary re-initialization.
:star: Added strict equality check

### DIFF
--- a/src/parsers/utils/areConnected.ts
+++ b/src/parsers/utils/areConnected.ts
@@ -29,5 +29,6 @@ export function areConnected(atom1: Atom, atom2: Atom) {
   )
     return false;
 
+    
   return true;
 }
--- a/src/parsers/utils/areConnected.ts
+++ b/src/parsers/utils/areConnected.ts
@@ -1,34 +1,33 @@
+import { Atom } from "./assignBackboneHBonds";
 import { bondLength } from "./bondLength";
 
 /*
  * Return true if atom1 and atom2 are probably bonded to each other based on distance alone
-*/
-
-export function areConnected(atom1: { elem: any; x: number; y: number; z: number; altLoc: string; }, atom2: { elem: any; x: number; y: number; z: number; altLoc: string; }) {
-  var maxsq = bondLength(atom1.elem) + bondLength(atom2.elem);
+ */
+export function areConnected(atom1: Atom, atom2: Atom) {
+  let maxsq = bondLength(atom1.elem) + bondLength(atom2.elem);
   maxsq += 0.25; // fudge factor, especially important for md frames, also see 1i3d
   maxsq *= maxsq;
 
-  var xdiff = atom1.x - atom2.x;
-  xdiff *= xdiff;
+  let xdiff = atom1.x - atom2.x;
+  xdiff = Math.pow(xdiff, 2);
   if (xdiff > maxsq) return false;
-  var ydiff = atom1.y - atom2.y;
+  let ydiff = atom1.y - atom2.y;
   ydiff *= ydiff;
   if (ydiff > maxsq) return false;
-  var zdiff = atom1.z - atom2.z;
+  let zdiff = atom1.z - atom2.z;
   zdiff *= zdiff;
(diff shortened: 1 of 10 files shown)


prajwalkulkarni:
I aim to improve the type-safety across the complete codebase in iterations. In this PR, the type of a few properties in `Atom` interface are still defined as `any`(and at few a places the types are also asserted as `any`, e.g: In `MMTF.ts`) as I'm still wrapping my head around the codebase and the type of the properties will be updated in further iterations.

8/13 files of `parsers/utils` are refactored in this PR, kindly confirm if I should push the refactored changes of the remaining 5 files of `parsers/utils` in this PR or create a separate PR. Thanks!

src/parsers/utils/assignBackboneHBonds.ts
```diff
@@ -1,15 +1,47 @@
 // This will identify all hydrogen bonds between backbone
 // atoms; assume atom names are correct, only identifies
 // single closest hbond
-export function assignBackboneHBonds(atomsarray: string | any[], hbondCutoff: number) {
-  let maxlength = hbondCutoff || 3.2;
-  let maxlengthSq = maxlength*maxlength;
-  let atoms = [];
+// interface Atoms {index: number; atom: string; hbondDistanceSq: number; hbondOther: any; hetflag:any}
+
+export interface Atom {
```
dkoes, line 6: AtomSpec is already defined in spec.ts.  Please do not define a duplicate type.

src/parsers/utils/areConnected.ts
```diff
@@ -1,34 +1,34 @@
...
+ */
+export function areConnected(atom1: Atom, atom2: Atom) {
+  let maxsq = bondLength(atom1.elem) + bondLength(atom2.elem);
   maxsq += 0.25; // fudge factor, especially important for md frames, also see 1i3d
   maxsq *= maxsq;
 
-  var xdiff = atom1.x - atom2.x;
-  xdiff *= xdiff;
+  let xdiff = atom1.x - atom2.x;
+  xdiff = Math.pow(xdiff, 2);
```
dkoes, line 13: No, pow is less efficient than multiply.

src/parsers/utils/atomNameToElem.ts
```diff
@@ -2,23 +2,31 @@ import { bondTable } from "./bondLength";
...
-      } else if(nothetero) {
-          if(elem == 'Ca') { //alpha carbon, not calcium
-              elem = 'C';
-          }
-          else if(elem == 'Cd') {
-              elem = 'C';
-          }
+  if (elem.length > 1) {
+    elem = elem[0].toUpperCase() + elem.substring(1).toLowerCase();
+    if (typeof bondTable[elem] === "undefined") {
```
dkoes, line 19: why not directly compare to undefined? The typeof comparison is left over from javascript where undefined was not a reserved keyword (not true in TypeScript).

dkoes (changes requested):

src/parsers/utils/assignBackboneHBonds.ts
prajwalkulkarni: Thank you pointing it out, I was not aware of it mostly because it was not used anywhere in `utils`.
Additionally, I have also added some of the missing types in `AtomSpec`. Currently, all the properties are optional, which shouldn't be the case, but as I further refactor other files and get better types for defining the properties, I'll update them in subsequent commits/PRs

src/parsers/utils/atomNameToElem.ts
dkoes: I forgot the other reason to use typeof is it doesn't throw an error if the thing doesn't exist.  So this should be replaced with direct comparisons to undefined with care.

dkoes (approved):

PR #761, opened 2024-01-26
Refactor remaining files in parser/utils

prajwalkulkarni:
# :dizzy: Changelog
:star: Improved typesafety in the remaining TS files located in `parser/utils`.
:star: Changed `var` to `let` and `const` as applicable.
:star: Added strict equality check
:star: If the name of identifiers of key and value are the same, do not pass the value identifier explicitly. E.g: Instead of using `atoms.push({x:x,y:y})` simply use `atoms.push({x,y})`


dkoes (approved):
Thanks!

PR #796, opened 2024-07-17
remove leftover debug code: console.log() calls

NicolasCARPi:
Hello,

After upgrading to 2.2.1, I realized my console was polluted by what appears to be leftover debug code:

2024-07-18-005125_1106x162_scrot

I commented out the culprit lines. Also removed some whitespace and added a spacing after an if.

My recommendations:

- use `console.err()` or `console.warn()` when you wish to log an error or a warning
- actually, don't use directly `console` but have a generic function such as `log(level, msg)` where you provide a level (debug, notice, warn, err) and the wrapper will only emit a message in the console if the loglevel is below the message level. So in dev you'll have your loglevel to "show me everything", and in prod it will just show the warn and errors. This will allow you to leave the debug lines in your code instead of having to comment them out, which is prone to errors.
- use a linter that will warn you if `console.log()` calls are present (https://eslint.org/docs/latest/rules/no-console)
- same linter can also automatically fix trailing whitespaces, incorrect indent or incorrect spacing for expressions

Best,
~Nicolas

### DIFF
--- a/src/GLViewer.ts
+++ b/src/GLViewer.ts
@@ -664,7 +664,7 @@ export class GLViewer {
             this.intwatcher = new window.IntersectionObserver(intcallback);
             this.intwatcher.observe(this.container);
         }
-          
+
         try {
             if (typeof (this.callback) === "function")
                 this.callback(this);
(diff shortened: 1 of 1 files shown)


dkoes:
This is taken care of in 2.2.1-post1

NicolasCARPi:
> This is taken care of in 2.2.1-post1

It seems it is still an issue in 2.3.0?

2024-08-04-173807_206x166_scrot

PR #797, opened 2024-07-20
fix: parse error for atomSymbol separate with one space

Rasic2:
If the atomSymbol in POSCAR or XDATCAR separated with only one space, like this:
```text
Ce O  C  O
```
The atomSymbol will return ["CeO", "C", "O"], and raise the warning: "declaration of atomary species wrong:"

This PR fix this by replace the code below

```js
  var atomSymbols = lines[5]
    .replace(/\s+/, "")
    .replace(/\s+$/, "")
    .split(/\s+/);
```
with this
```js
  var atomSymbols = lines[5]
    .replace(/\s*/, "")
    .replace(/\s+$/, "")
    .split(/\s+/);
```


dkoes:
I am not familiar with these formats.  Why is the this the right fix?  Changing from `\s+` to `\s*` will only remove space if it is at the very beginning.  If this is what is intended, then why not use `.trim` to remove leading and trailing white space?

dkoes:
This is addressed in d9efe75

PR #814, opened 2024-11-12
fix docstring CIF location

cmishra:
minor fix so `CIF` shows back up in the documentation instead of the helper function `fractionalToCartesian`


dkoes:
Thanks!

Issue #704, opened 2023-07-17
Support for ES6-loadable bundles?


arogozhnikov:
I'm using ES6 bundles, but I do not develop with npm, just import them in a browser.
That's not extremely common scenario, but a very useful one:

1. just editor and browser - no npm, building, source mapping, bundling, etc. Correspondingly, no need to care about yearly updates of anything in the node world
2. at the same time imports are isolated, so e.g. $ and $3dmol do not load into global space
3.  ~~and a sweet part - autosuggestions work~~ (seems they do not work for 3dmol)

I suggest serving ES6 module for 3Dmol, can help with bundling code.

Update: analysis works only partially, e.g. suggest deprecated elements, and can jump to implementation:

<img width="654" alt="Screenshot 2023-07-16 at 10 39 19 PM" src="https://github.com/3dmol/3Dmol.js/assets/6318811/cde79ef4-d1f9-4021-926d-f6c8af6c6718">

Support for other exported packages is usually better, but even in this state it is way more helpful compared to <script>  inclusion.

dkoes:
I am confused about what you are asking for.  We target UMD with webpack for maximum compatibility.  $ does not get exported to the global namespace, but $3Dmol will for backwards compatibility.

If you have a suggestion for how to change the webpack configuration to better match your workflow **without breaking backwards compatibility** I am happy to consider it.

arogozhnikov:
That's how it works for me now:
```
import { $3Dmol } from "https://arogozhnikov.github.io/es6_importable_packages/bundles/3dmol.js";
```
(you can try that too).

I can use it straight from JS, and then open page right away in the browser. 
No npm, no node, no bundler. Code analysis half-works. 

On the [official page](https://3dmol.csb.pitt.edu/doc/index.html) I see only <script> tag option for loading.

dkoes:
Glad you got it to work.  As far as I can tell, ES6 export is still in development in webpack (and has been for 7 years - https://github.com/webpack/webpack/issues/2933) so it isn't clear to me how to implement this in our build system and in a backward compatible manner.

arogozhnikov:
I use rollup.

AFAIK (not an expert) es6 modules are not compatible with other formats (and that's the), so this is basically +1 exporting pipeline (which produces separate 3dmol.esm.js and 3dmol.esm-min.js). 

I'm ready to work on PR that ships bundles if you think that's valuable. Promise not to break your existing webpack pipeline :)

dkoes:
[...]

PR #831, opened 2025-05-07
Build as both UMD and ES6 modules.

asford:
Ref: #704 

Updates 3Dmol to build both UMD and ES6 compatible modules, allowing imports via `<script type="module"/>` in web contexts.

This includes...

... lightly refactoring the webpack configuration to expose UMD/ES6 and development/production targets.
... lightly refactoring the module layout to expose a namespace target that does not bind `window.$3Dmol`.
... `colors.ts` fixed to explicitly reference htmlColors, rather than `$3Dmol`.
... falling back to sync surface rendering in ES6 configs. The surface rendering webworker source is ... trickly ... to build in the ES6 configuration. It *could* be ported to a module worker, but this is out of scope for this work.

This should be fully backward compatible with previous build targets.
As an example smoketest (likely to be reverted before merge), this updates the top-level `index.html` to use the ES6 build.


asford:
@dkoes would you mind taking a look at this config change and assess it for a merge?
If you're 👍 on merging we can make some lightweight documentation and example updates to include both the ES6 and UMD style imports.

dkoes (approved):
Looks good to me - would appreciate extra documentation.

Issue #455, opened 2020-05-01
Aromatic bond visualization


dkoes:
Stick style should show aromatic bonds with solid/dashed representation.

dxdc:
@dkoes 

I've done more follow up on #639, and I have a working implementation that could also be used for this feature (have not pushed the PR yet). It's a more general solution in which one can specify partial bond order of any bond to control the thickness of the remaining dashed bond.

For example, instead of specifying a bond order of 2, it uses 1.7 in this example:

<img width="400" alt="Screenshot 2024-01-10 at 9 24 47 AM" src="https://github.com/3dmol/3Dmol.js/assets/7200365/6ffaf1d2-00f3-4cb4-8ad1-18027a529d4f">

--

However, that brings up a few questions.

1. Is there a mathematical way to determine -- in a multi-bond situation -- which of the bonds should be dashed? Consider this example, with the dashed bonds reversed.

<img width="400" alt="image" src="https://github.com/3dmol/3Dmol.js/assets/7200365/006cab1f-625b-4af5-acad-8aee62a768e8">

In my WIP implementation, I have done it in two ways, but a more mathematical default would be superior (e.g., to somehow prioritize the "shorter" bond as dashed).

- `dashPriority` property on the atom itself
- Bond order which ends with `001`, e.g. `1.7001`. This could be easier, in some cases, for tuning SDF files (particularly in the context of an animation) without trying to properly determine which atom it should be assigned to in the settings.

```js
viewer.setStyle({serial: 4}, {
    stick: {
        radius: 0.2,
        dashPriority: true,
    }
});
```

2. In cases where atoms have different colors, what is the correct way to determine the dash color?

- Existing implementation (about half and half)
<img width="400" alt="Screenshot 2024-01-09 at 8 11 24 PM" src="https://github.com/3dmol/3Dmol.js/assets/7200365/fb02e2b9-d3e5-4522-8875-82ac87b4ec22">

- Gradient color scheme, slowly transitioning from one to the other
<img width="400" alt="Screenshot 2024-01-09 at 8 05 26 PM" src="https://github.com/3dmol/3Dmol.js/assets/7200365/9fdb7a1a-1339-45c9-b5df-530672d2f477">

- Calculate the most appropriate color by relative contrast, or another means
[...]

PR #842, opened 2025-12-31
Aromatic/dashed bond order rendering #455

dxdc:
This PR extends the dashed bond work from #641 and #740 to handle aromatic and fractional bond orders. Bonds with order 1.5, 2.5, or type 4 (aromatic) now display as a solid+dashed pair matching standard chemical drawing conventions.

### Background

PR #641/#740 added dashed bond rendering for bond orders < 1. This PR builds on that foundation to handle the more complex case raised in #455: aromatic bonds (type 4 in SDF) and fractional orders like 1.5 need to render as a solid+dashed pair, not just a single dashed line. The tricky part is determining which side gets the dashed portion - it should face the ring interior.

### What changed

- Fractional bond orders (1.5, 2.5) and aromatic type (4) render as one solid cylinder plus one dashed cylinder. The dashed portion automatically faces the ring interior by computing a weighted centroid of neighboring atoms. Ring atoms get full weight, terminal hydrogens get 0.25 weight.

- Fine-tuned previous formula for dash creation

- Two-color bonds between different atom types (e.g. C-N in pyridine) now split correctly at the midpoint for both solid and dashed portions.

- Added solidColor and dashedColor properties to dashedBondConfig for explicit color control: 
```js
{stick: {dashedBondConfig: {dashLength: 0.1, gapLength: 0.2, solidColor: '#00F', dashedColor: '#F80'}}}
```

- Also fixed getSideBondV to handle edge cases better when finding neighbor atoms for bond offset direction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "jest": "npx jest --testPathIgnorePatterns=render.test.js",
     "test": "npm run generate:jest && npm run jest && npm run glcheck",
     "glcheck": "npm run clean:glcheck && npm run generate:glcheck && npm run test:glcheck",
-    "test:glcheck": "npx glcheck --config ./tests/glcheck/glcheck.config.json --capture-missing true",
+    "test:glcheck": "npx glcheck --config ./tests/glcheck/glcheck.config.json",
     "clean:glcheck": "rm -rf ./tests/glcheck/render-failures",
     "generate:tests": "python3 tests/auto/generate_tests.py",
     "generate:glcheck": "python3 ./tests/glcheck/generate_glcheck_render_tests.py -min",


package.json
```diff
@@ -38,7 +38,7 @@
     "jest": "npx jest --testPathIgnorePatterns=render.test.js",
     "test": "npm run generate:jest && npm run jest && npm run glcheck",
     "glcheck": "npm run clean:glcheck && npm run generate:glcheck && npm run test:glcheck",
-    "test:glcheck": "npx glcheck --config ./tests/glcheck/glcheck.config.json",
+    "test:glcheck": "npx glcheck --config ./tests/glcheck/glcheck.config.json --capture-missing true",
```
dkoes: Please undo this change - this command should only be run with capture-missing true when you are generating the reference image.

dxdc:
thanks @dkoes! any chance of pushing this into a new release? :)

PR #845, opened 2026-01-20
Fix dashed bond edge cases and stabilize dash generation

dxdc:
- Fixes calculateDashes() producing NaN/Infinity for short bonds when gap length collapses.
- Stabilizes dashed bond rendering for fractional/aromatic bond orders.
- No intended visual changes for normal bonds; prevents corrupted geometry in edge cases.

### DIFF
--- a/src/GLModel.ts
+++ b/src/GLModel.ts
@@ -968,8 +968,8 @@ export class GLModel {
                         toCap = 2;
 
                     const isDashed = atomDashedBonds || (renderBondOrder % 1 !== 0);
-                    const colA = isDashed ? dashed1 : solid1;
-                    const colB = isDashed ? dashed2 : solid2;
+                    const colA = isDashed ? dashed1 : C1;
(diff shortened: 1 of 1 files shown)


src/GLModel.ts
```diff
@@ -952,7 +967,11 @@ export class GLModel {
                     if (!atom2.capDrawn && atom2.bonds.length < 4)
                         toCap = 2;
 
-                    drawCyl(geo, p1, p2, bondR, dashed1, dashed2, fromCap, toCap);
+                    const isDashed = atomDashedBonds || (renderBondOrder % 1 !== 0);
+                    const colA = isDashed ? dashed1 : solid1;
```
dkoes: If solidColor is set, it will also be used to color single bonds and renderBondOrder > 3 bonds.  Is this what is intended? I think it should be C1/C2 unless renderBondOrder < 1

partialbonds4

